PR #230, opened 2015-01-20
Handle 'observation message packing: integer overflow' more gracefully

fnoble:
This error does happen very occasionally as reported here:
https://groups.google.com/d/msg/swiftnav-discuss/8bdngdOkXKM/lz_Ov7PEL7wJ

This error should be non-fatal. This patch just reports the error and drops the offending satellite from the observation message.

### DIFF
--- a/src/sbp_piksi.h
+++ b/src/sbp_piksi.h
@@ -160,7 +160,7 @@ void pack_obs_header(gps_time_t *t, u8 total, u8 count,
 void unpack_obs_content(msg_obs_content_t *msg,
   double *P, double *L, double *snr, u16 *lock_counter, u8 *prn);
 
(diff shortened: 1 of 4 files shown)


src/sbp_piksi.h
```diff
@@ -160,7 +160,7 @@ void pack_obs_header(gps_time_t *t, u8 total, u8 count,
 void unpack_obs_content(msg_obs_content_t *msg,
   double *P, double *L, double *snr, u16 *lock_counter, u8 *prn);
 
-void pack_obs_content(double P, double L, double snr, u16 lock_counter, u8 prn,
```
gsmcmullin: Maybe return bool and/or document return value.

PR #231, opened 2015-01-20
Handle RX DMA buffer overflow more gracefully

### LINKED ISSUE #218: DMA buffer overrun should be non-fatal

fnoble:
Buffer overflow error is now non-fatal.

Add 'IO error' count functionality, reported by the UART state message and in the console system monitor view.

Fixes #218


gsmcmullin:
This looks good and works for me.

PR #239, opened 2015-01-29
Console tweaks for new C/N0 units

fnoble:
* Fix error with tracking view caused by new C/N0 units
 * Update label in observation view


cbeighley:
lgtm

PR #240, opened 2015-01-30
Observation handling fixes

### LINKED ISSUE #211: base_obss semaphore flagged from the wrong place

fnoble:
Fixes #211 

The second commit in the PR fixes the issue of not going into float mode immediately upon getting >= 5 sats. The problem was the observations did not contain a correct Doppler which is used to calculate the velocity and hence triggered the position solver to filter out due to the crazy velocity. This patch uses the `tdcp_doppler` function to fill in the Doppler based on the time difference of carrier phase observables.

The rest is mainly refactoring. I pulled the base station observation handling code into its own file, cleaned it up and added much more thorough documentation.

### DIFF
--- a/src/base_obs.c
+++ b/src/base_obs.c
@@ -128,9 +128,11 @@ static void update_obss(obss_t *new_obss)
   /* Reset the `has_pos` flag. */
   u8 has_pos_old = base_obss.has_pos;
   base_obss.has_pos = 0;
-  /* Check if the base station has sent us its position explicitly and if so us
+  /* Check if the base station has sent us its position explicitly via a
+   * BASE_POS SBP message (as indicated by #base_pos_known), and if so use
    * that. No need to lock before reading here as base_pos_* is only written
-   * from this thread (SBP). */
+   * from this thread (SBP).
+   */
   if (base_pos_known) {
     /* Copy the known base station position into `base_obss`. */
     memcpy(base_obss.pos_ecef, base_pos_ecef, sizeof(base_pos_ecef));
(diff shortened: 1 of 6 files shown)


src/base_obs.c
```diff
@@ -0,0 +1,329 @@
...
+  chMtxLock(&base_pos_lock);
+  wgsllh2ecef(llh, base_pos_ecef);
+  base_pos_known = true;
+  chMtxUnlock();
+}
+
+/** SBP callback for the old style observation messages.
+ * Just prints a deprecation warning. */
+static void obs_old_callback(u16 sender_id, u8 len, u8 msg[], void* context)
+{
```
imh: should this do something? maybe deprecate it but leave it functional for a while?

fnoble: You mean obs_old_callback? It would be hard to leave it functional but its nice to at least print a warning for a while. Its been this way for quite a while now.

src/base_obs.c
```diff
@@ -0,0 +1,329 @@
...
+         new_obss->n * sizeof(navigation_measurement_t));
+  n_old = new_obss->n;
+
+  /* Reset the `has_pos` flag. */
+  u8 has_pos_old = base_obss.has_pos;
+  base_obss.has_pos = 0;
+  /* Check if the base station has sent us its position explicitly and if so us
+   * that. No need to lock before reading here as base_pos_* is only written
+   * from this thread (SBP). */
+  if (base_pos_known) {
```
imh: where is the sent position coming from? If it moves around (e.g. single point position), it will kill the low latency mode.

src/base_obs.c
```diff
@@ -0,0 +1,329 @@
...
+    /* Indicate that the position is valid. */
+    base_obss.has_pos = 1;
+  /* The base station wasn't sent to us explicitly but if we have >= 4
+   * satellites we can calculate it ourselves (approximately). */
+  } else if (base_obss.n >= 4) {
+    gnss_solution soln;
+    dops_t dops;
+
+    /* Calculate a position solution. */
+    s32 ret = calc_PVT(base_obss.n, base_obss.nm, &soln, &dops);
```
imh: this is the part that was locking us out from the weird dopplers before?

src/base_obs.c
```diff
@@ -0,0 +1,329 @@
...
+{
+  (void) context;
+
+  static s16 prev_count = 0;
+  static gps_time_t prev_t = {.tow = 0.0, .wn = 0};
+  /* Using an extra obss_t so we don't need to overwrite the old set (used for
+   * low latency) when we end up with a bad new set. */
+  static obss_t base_obss_raw = {.has_pos = 0};
+
+  /* Sender ID of zero means that the messages are relayed observations,
```
imh: should it be clear what this is referring to?

src/base_obs.c
```diff
@@ -0,0 +1,329 @@
...
+
+  /* Calculate packet latency. */
+  if (time_quality >= TIME_COARSE) {
+    gps_time_t now = get_current_time();
+    float latency_ms = (float) ((now.tow - t.tow) * 1000.0);
+    log_obs_latency(latency_ms);
+  }
+
+  /* Verify sequence integrity */
+  if (count == 0) {
```
imh: the semantics of count and prev_count are still unclear. Can we document what we're counting?

src/base_obs.c
fnoble: This is if the base station has an accurate surveyed location that it sends out, its based on user input and presumed to be correct. Not really any way to verify the user hasn't screwed up so we kinda have to trust them.

src/base_obs.c
fnoble: yes

src/base_obs.c
fnoble: not sure which line you are referring to?

src/base_obs.c
fnoble: ok, will do

src/base_obs.c
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  /* Pull out the contents of the message. */
+  msg_obs_content_t *obs = (msg_obs_content_t *)(msg + sizeof(msg_obs_header_t));
+  for (u8 i=0; i<obs_in_msg; i++) {
+    /* Check if we have an ephemeris for this satellite, we will need this to
+     * fill in satellite position etc. parameters. */
+    if (ephemeris_good(es[obs[i].prn], t)) {
+      /* Unpack the observation into a navigation_measurement_t. */
+      unpack_obs_content(
```
imh: why have an unpack function that unpacks into a bunch of random addresses instead of a message --> navigation_measurement_t?

src/base_obs.c
```diff
@@ -0,0 +1,329 @@
...
+ */
+static void obs_callback(u16 sender_id, u8 len, u8 msg[], void* context)
+{
+  (void) context;
+
+  static s16 prev_count = 0;
+  static gps_time_t prev_t = {.tow = 0.0, .wn = 0};
+  /* Using an extra obss_t so we don't need to overwrite the old set (used for
+   * low latency) when we end up with a bad new set. */
+  static obss_t base_obss_raw = {.has_pos = 0};
```
imh: We need a better name for this

src/base_obs.c
fnoble: Because Niels did it that way initially, I guess it makes it more general purpose and less libswiftnav specific. I could change it but it might be out of the scope of this PR?

PR #245, opened 2015-02-06
Adding ARM to the list of reasons to rename strnicmp to the posix strncasecmp

soulcmdc:
This fixes the current issue with the compilation on Ubuntu 14.10/14.04


fnoble:
Looks like this function was removed from newlib recently, probably means this will apply to other platforms soon when they update their packages.

PR #248, opened 2015-02-14
Prioritizing the search for viable/visible common view satellites

fnoble:


fnoble:
@cturvey: I rebased your changes off of the current master and created this pull request to code review.

One thought - what would happen if there was a satellite in the base station observation set that was not visible to the rover? If the observations were received faster than the acquisition could complete then there is potential for a loop where the PRN gets set back to untried before any other PRNs get a chance to run. Perhaps we need to limit how often we "prod"?

cturvey:
I'm still getting my head around what's happening inside the firmware. I
don't think this generates a race condition, and most side-effects should
be relatively benign, especially if you increase the number of
correlators/trackers so there are always free resources. The corner
condition I see is where you're tracking 10 satellites, and there are
perhaps one or two others that might be better candidates. My immediate
thought would be to score based on CNO/SNR, or perhaps get the Elevation
stuff working. The signal strength is by far the cheapest, and a data point
provided in the SBP OBS packet.

There is a stronger case in my mind to rework the whole acquisition
management, where the "what's in view" stuff, and a lot of the other over
visited work, gets decimated and done far less frequently. For example
generating a bit vector for in-view, and healthy satellites that gets
checked rather than doing the math over and over.

The new FPGA might address some of the sensitivity issues I'm seeing, I
don't have much insight into that. I see people on the forum who have a lot
of tracked channels, but this hasn't been my experience, I tend to see 5 or
6 satellites, the base and rover see different ones, and the receivers
spend a lot of time looking for the wrong satellites, and don't find the
ones that should be there. This is with some very good antennas, with good
placement, in the flat open plains of Illinois. The mutual sharing of
satellite I can see between base and rover has been very effective here at
focusing effort and getting us to 9 to 10 common view solutions.

Now I have the build process down, and some stable firmware and RTK fixing,
I'll dig into this some more.

-Clive

fnoble:
@cturvey : Just wanted to let you know I hadn't forgotten about your patch!

I really want to get this change into the next release but still believe that there is a chance the acquisition could get tied up looking for a satellite that isn't there but keeps getting prodded. In the worst case if the prodding was faster than the tim the acquisition takes to do a search then no other satellites will be attempted.

Agree that the whole acquisition management logic could be refactored and improved. Will think about how to address these issues and welcome any feedback.

cturvey:
I guess my point is that the other receiver can see the satellite(s) I'm
asking this one to look for. The satellite might be obscured, or the other
receiver is so far away it's over the horizon, but it's not behind the
planet.

You could weight on elevation or snr/cno? The strategy could change if it's
apparent the baseline >35km.

Ideally if you know your time and approximate location, or already have
four satellite, and almanac you can figure whats in the sky and eliminate
half the constellation. You don't need to do this with high frequency,
every 15 seconds would be just fine. Create a bit vector of visible/healthy
satellites, and when the receiver isn't in a sky-search mode, it only looks
for satellites described in the bit vector. This vector could be added too
by the other receivers visible satellites, but realistically they are going
to be pretty much identical in a working system.

What I would try to avoid is a lot of unnecessary math. ie appropriate to
the rate of change, not every time through the loop.

Perhaps the FPGA can get more tracking channels, 14 would be ideal, 12 was
the target when the full constellation was supposed to have 24 satellites.

gsmcmullin:
Included in #303.

PR #254, opened 2015-03-03
Use ANSI sequence to clear line.

gsmcmullin:


fnoble:
Will this work on Windows? (maybe we don't care)

What are the benefits of doing it this was?

gsmcmullin:
It spewed unprintable binary for me on Linux.
I haven't tested on Windows.

fnoble:
@cbeighley : any comments? the fix looks good to me but presumably this worked in some context before?

gsmcmullin:
It doesn't work in Windows, but Windows doesn't get this far.  It aborts in serial_link.py if the port doesn't exist.  The old code doesn't work in Windows either, chr(3) prints a heart.  I guess it worked on OS X before.

fnoble:
lol, ok, wtf is 0x03 "End of text" anyhow?

PR #256, opened 2015-03-05
Updates pygments dependency in setup script.

mookerji:
The listed pygments requirement in the provisioning script for OS X
was apparently outdated.

r=@fnoble,@cbeighley,@denniszollo.


fnoble:
There is another way we can do this, the requirement for this module was only created for packaging on Windows with the latest version of pygments. The modules isn't actually used so we can just wrap it in a try block.

fnoble:
This is the offending line:

https://github.com/swift-nav/piksi_firmware/blob/master/scripts/console.py#L65

Purely imported so that pyinstaller can pick it up, not actually used for anything.

mookerji:
@fnoble and I discussed offline, and I think this is good to go now.

PR #263, opened 2015-03-09
Fixing case when amb test tries to use too many sats (DONT MERGE)

imh:
Also: 
- made a bunch of stuff const
- printing number of sats in projection like we do in inclusion.
- other, to be documented in the real PR


mookerji:
retest this please

PR #266, opened 2015-03-12
Updated default values and logical units of simulator noise settings.

denniszollo:
Now when the setting is called sigma it is an actual standard deviation not a variance.
Default sim settings were changed slightly to acheive round numbers

### DIFF
--- a/src/simulator.c
+++ b/src/simulator.c
@@ -46,11 +46,11 @@ simulation_settings_t sim_settings = {
   },
   .speed = 4.0,
   .radius = 100.0,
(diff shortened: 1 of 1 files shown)


src/simulator.c
```diff
@@ -44,13 +44,13 @@ simulation_settings_t sim_settings = {
     -4292474.39651309,
     3855434.34087421
   },
-  .speed = 4.0,
+  .speed = 4.0,
```
fnoble: Trailing whitespace

PR #267, opened 2015-03-16
Use new libswiftnav logging infrastructure

fnoble:
Replace `printf()` with calls to `log_error`, `log_warn` and `log_info` to annotate different messages with a log level.

For now these macros just map back to `printf` with a prefix to identify the log level but now the various messages going through the logging calls its easy to change the handling of different message levels on the backend.

\cc: @gsmcmullin @denniszollo @mookerji

### DIFF
--- a/src/acq.c
+++ b/src/acq.c
@@ -12,9 +12,10 @@
 
 #include <math.h>
 #include <string.h>
-#include <stdio.h>
 
 #include <ch.h>
+
+#include <libswiftnav/logging.h>
 
 #include "board/nap/acq_channel.h"
 #include "acq.h"
@@ -32,8 +33,9 @@ void acq_set_prn(u8 prn)
 {
   chBSemInit(&load_wait_sem, TRUE);
   nap_acq_code_wr_blocking(prn);
-  if (chBSemWaitTimeout(&load_wait_sem, 1000) == RDY_TIMEOUT)
-    printf("acq: Timeout waiting for code load!\n");
+  if (chBSemWaitTimeout(&load_wait_sem, 1000) == RDY_TIMEOUT) {
+    log_warn("acq: Timeout waiting for code load!\n");
+  }
 }
 
 /** Send results of an acquisition to the host.
@@ -75,7 +77,7 @@ bool acq_load(u32 count)
   nap_acq_load_wr_enable_blocking();
   nap_timing_strobe(count);
   if (chBSemWaitTimeout(&load_wait_sem, 1000) == RDY_TIMEOUT) {
-    printf("acq: Timeout waiting for sample load!\n");
+    log_warn("acq: Timeout waiting for sample load!\n");
     return false;
   }
   return true;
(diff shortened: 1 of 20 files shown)


src/acq.c
```diff
@@ -75,7 +77,7 @@ bool acq_load(u32 count)
   nap_acq_load_wr_enable_blocking();
   nap_timing_strobe(count);
   if (chBSemWaitTimeout(&load_wait_sem, 1000) == RDY_TIMEOUT) {
-    printf("acq: Timeout waiting for sample load!\n");
+    log_warn("acq: Timeout waiting for sample load!\n");
```
denniszollo: Do we have the capacity to put more info in the messages.  For instance, how long was the timout parameter here?

fnoble: Yeah, the `log_*` functions work just like `printf` so we can add any info we want. Do you want me to add the timeout length to these messages? Its fairly low level and the timeout isn't configurable by the user.

cbeighley:
lgtm

src/acq.c
denniszollo: Not for this one, but maybe there are some opportunities to make the messages more informative so users or integrators wouldn't need to go to the source in order to understand.

src/base_obs.c
```diff
@@ -237,7 +237,7 @@ static void obs_callback(u16 sender_id, u8 len, u8 msg[], void* context)
    * i.e. is it going to time match one of our local obs. */
   double epoch_count = t.tow * (soln_freq / obs_output_divisor);
   if (fabs(epoch_count - round(epoch_count)) > TIME_MATCH_THRESHOLD) {
-    printf("Unaligned observation from base station ignored.\n");
+    log_warn("Unaligned observation from base station ignored.\n");
```
denniszollo: Perhaps we could put the tow at which this occurred for this message?

src/peripherals/usart.c
```diff
@@ -183,9 +184,9 @@ void usarts_enable(u32 ftdi_baud, u32 uarta_baud, u32 uartb_baud, bool do_precon
 
   if (do_preconfigure_hooks) {
 
-    printf("\n\nPiksi Starting...\n"
-       "Firmware Version: " GIT_VERSION "\n" \
-       "Built: " __DATE__ " " __TIME__ "\n");
+    log_info("Piksi Starting...\n");
```
denniszollo: I kind of like the idea of this being a log_warn.  It's nice to have a message on startup from a usability perspective and with this change I don't think anyone will see the Piksi starting and the firmware version.

fnoble: I feel like log_warn should be reserved for bad things only, it makes it easier for people to react to the warnings if they know there are none that should just be ignored. There are a couple options:

 - Ship with log level set to INFO
 - Don't worry about it because the same info can be found on the settings tab
 - Have the console detect the `MSG_STARTUP` message and print or show some other kind of indication of the Piksi booting up

src/base_obs.c
fnoble: Will do!

src/peripherals/usart.c
denniszollo: I like option 3.

src/acq.c
mookerji: +1.

src/acq.c
```diff
@@ -32,8 +33,9 @@ void acq_set_prn(u8 prn)
 {
   chBSemInit(&load_wait_sem, TRUE);
   nap_acq_code_wr_blocking(prn);
-  if (chBSemWaitTimeout(&load_wait_sem, 1000) == RDY_TIMEOUT)
-    printf("acq: Timeout waiting for code load!\n");
+  if (chBSemWaitTimeout(&load_wait_sem, 1000) == RDY_TIMEOUT) {
```
mookerji: nit/lame: it's unrelated to this patchset, but it would be cool to replace the timeout values with a macro reference.

src/main.c
```diff
@@ -87,18 +87,18 @@ static msg_t nav_msg_thread(void *arg)
         __asm__("CPSIE i;");
 
         if (ret < 0) {
-          printf("PRN %02d ret %d\n", tracking_channel[i].prn+1, ret);
+          log_info("PRN %02d ret %d\n", tracking_channel[i].prn+1, ret);
```
mookerji: Yeah, so I remember seeing this before in the logs and being confused. Would it be possible to add some additional context here, such as the meaning of this return value and context of the statement?

src/manage.c
```diff
@@ -177,7 +178,7 @@ u8 best_prn(void)
       if (acq_prn_param[prn].state == ACQ_PRN_TRIED)
         acq_prn_param[prn].state = ACQ_PRN_UNTRIED;
     }
-    printf("acq: restarting PRN search\n");
+    log_info("acq: restarting PRN search\n");
```
mookerji: so, I do find it useful that these messages are prefixed by the subsystem of origin. I think something similar could be done with other `log_info`s here (such as the `log_info("PRN ... unhealthy");` line, for example.

mookerji:
Sweet.

src/manage.c
gsmcmullin: It may be useful to push such a log tag up as a separate string, so the console or other log inspection tool could filter by tag/subsystem.

src/base_obs.c
fnoble: Done, but will submit as a separate PR

src/manage.c
fnoble: I think this is a good idea but I might follow up with those changes in a separate PR

fnoble:
Going to merge and address some of the comments with additional PRs because some of it is out of the scope of what this particular change was trying to achieve.

PR #268, opened 2015-03-17
serial_link watchdog

mfine:
Take a parameter on the command line to configure a watchdog timer: if a heartbeat is not heard in `X` seconds, exit the process. Useful to drive long running health tests. Could eventually have a deeper health check here besides just simple heartbeats. Potentially there's scope for that in the new client with various configurable "watchdogs"?

/cc @fnoble @mookerji

### DIFF
--- a/scripts/serial_link.py
+++ b/scripts/serial_link.py
@@ -21,6 +21,8 @@ import calendar
 import socket
 import json
 
+from sbp_messages import SBP_HEARTBEAT
+
 DEFAULT_PORT = '/dev/ttyUSB0'
 DEFAULT_BAUD = 1000000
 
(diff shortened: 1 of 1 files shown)


fnoble:
lgtm

scripts/serial_link.py
```diff
@@ -426,8 +453,12 @@ def format_log_json_entry(t0, item):
...
     else:
       link.add_global_callback(default_log_callback(log_file))
-    if args.reset:
-      link.send_message(ids.RESET, '')
+  if args.reset:
+    link.send_message(ids.RESET, '')
+  # Setup watchdog
+  watchdog = args.watchdog[0]
+  if watchdog:
+    link.add_callback(0xFFFF, Watchdog(float(watchdog), default_watchdog))
```
mookerji: nit: magic value: `0xFFFF` => `ids.HEARTBEAT` (I think).

scripts/serial_link.py
```diff
@@ -382,6 +382,30 @@ def format_log_json_entry(t0, item):
...
+      self.timer.cancel()
+    self.timer = threading.Timer(self.timeout, self.alarm)
+    self.timer.start()
+
+  def alarm(self):
+    import thread
+    thread.interrupt_main()
+
+def default_watchdog():
+  print "ERROR: Watchdog expired!"
```
mookerji: nit: may want to use stderr instead: `import os; os.stderr.write('ERROR: Watchdog expired!`).

scripts/serial_link.py
```diff
@@ -382,6 +382,30 @@ def format_log_json_entry(t0, item):
   data = item.to_json_dict()
   return {'delta': delta, 'timestamp': timestamp, 'data': data}
 
+class Watchdog:
+  def __init__(self, timeout, alarm):
```
mookerji: nit: a brief description, possibly saying that it calls an alarm raising a `KeyboardInterruptException` (?) or something after timeout.

scripts/serial_link.py
```diff
@@ -382,6 +382,30 @@ def format_log_json_entry(t0, item):
...
+    self.alarm = alarm
+    self.timer = None
+
+  def __call__(self, *args):
+    self.call()
+
+  def call(self):
+    if self.timer:
+      self.timer.cancel()
+    self.timer = threading.Timer(self.timeout, self.alarm)
```
mookerji: Also, lame question, but does `self.timer` need to be on a daemon thread (i.e., `self.timer.daemon = True`?

scripts/serial_link.py
mfine: Ha, yeah had tried that, didn't work. Looked closed, and have to do this to bring it in:

```python
from sbp_messages import SBP_HEARTBEAT
```

Isn't the wildcard in `sbp_piksi.py` doing that? Python, how does it work

Done.

scripts/serial_link.py
mfine: Used `sys.stderr`, since there's already some usage for `sys.stdout`. Done.

scripts/serial_link.py
mfine: Done.

scripts/serial_link.py
mfine: Totally - forgot to set this, thanks for the catch! Done.

fnoble:
@mookerji are you happy to merge this now?

PR #270, opened 2015-03-17
Base observation handling tweaks

fnoble:
- Better warning message
 - Fixes small error with time matching logic
 - Filters out PRNs >31, fixes crash on receiving simulated observations

@denniszollo

### DIFF
--- a/src/base_obs.c
+++ b/src/base_obs.c
@@ -280,6 +280,11 @@ static void obs_callback(u16 sender_id, u8 len, u8 msg[], void* context)
   /* Pull out the contents of the message. */
   msg_obs_content_t *obs = (msg_obs_content_t *)(msg + sizeof(msg_obs_header_t));
   for (u8 i=0; i<obs_in_msg; i++) {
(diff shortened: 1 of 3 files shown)


src/base_obs.c
```diff
@@ -276,6 +280,11 @@ static void obs_callback(u16 sender_id, u8 len, u8 msg[], void* context)
   /* Pull out the contents of the message. */
   msg_obs_content_t *obs = (msg_obs_content_t *)(msg + sizeof(msg_obs_header_t));
   for (u8 i=0; i<obs_in_msg; i++) {
+    /* Check the PRN is valid. e.g. simulation mode outputs test observations
+     * with PRNs >200. */
+    if (obs[i].prn > 31) {
```
denniszollo: Perhaps we should fix simulator to output only PRNs that make sense?

fnoble: The simulator only outputs fake PRNs so as not to interfere with the filter.

PR #271, opened 2015-03-17
Simulator fixes

fnoble:
- Corrects time of simulated solution outputs
 - Respects the `obs_output_divisor` setting when sending simulated observations

@cbeighley @gsmcmullin

### DIFF
--- a/src/solution.c
+++ b/src/solution.c
@@ -491,7 +491,7 @@ static msg_t solution_thread(void *arg)
 
         u8 flags = simulation_enabled_for(SIMULATION_MODE_RTK) ? 1 : 0;
 
-        solution_send_baseline(&soln->time,
+        solution_send_baseline(&(soln->time),
           simulation_current_num_sats(),
           simulation_current_baseline_ecef(),
           simulation_ref_ecef(), flags);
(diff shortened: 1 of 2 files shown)


src/solution.c
```diff
@@ -467,36 +467,40 @@ static msg_t solution_thread(void *arg)
...
-      if ((simulation_enabled_for(SIMULATION_MODE_FLOAT) ||
-           simulation_enabled_for(SIMULATION_MODE_RTK)) &&
-          fabs(t_check - (u32)t_check) < TIME_MATCH_THRESHOLD) {
+      if (simulation_enabled_for(SIMULATION_MODE_FLOAT) ||
+          simulation_enabled_for(SIMULATION_MODE_RTK)) {
 
         u8 flags = simulation_enabled_for(SIMULATION_MODE_RTK) ? 1 : 0;
 
-        solution_send_baseline(&simulation_current_gnss_solution()->time,
+        solution_send_baseline(&soln->time,
```
denniszollo: I hate seeing this sort of thing because I can never remember c precedence rules. Do we ever put parenthesis in for clarity?  &(soln->time)

denniszollo:
Otherwise makes sense and looks good to me.  Timing of messages is important in integration testing so it's good that the user will see the round number he expects and the output divisor will be taken into account.

PR #273, opened 2015-03-18
Move on to libsbp

mfine:
First pass at moving on to libsbp. Moved messages and structs over. There are a couple of missing messages (todo: `OLD_OBS`, `PACKED_OBS`). Comments and feedback about how this is integrated (is there a better way to import messages and their structs?).

/cc @mookerji @fnoble


scripts/observation_view.py
```diff
@@ -24,7 +24,9 @@
 import numpy as np
 import datetime
 
-import sbp_piksi as ids
+#from sbp.deprecated import SBP_MSG_OLD_OBS TODO should we keep around deprecated messages?
```
mfine: Do we want to preserve deprecated messages and continue to support deprecated message paths like here?

scripts/acq_results.py
```diff
@@ -28,7 +30,7 @@ class AcqResults():
   def __init__(self, link):
     self.acqs = []
     self.link = link
-    self.link.add_callback(ids.ACQ_RESULT, self._receive_acq_result)
+    self.link.add_callback(SBP_MSG_ACQ_RESULT, self._receive_acq_result)
```
mookerji: Hmmm, _receive_acq_result should probably call the MsgAcqResult parser.

scripts/baseline_view.py
```diff
@@ -152,11 +153,11 @@ def update_table(self):
     self._table_list = self.table.items()
 
   def gps_time_callback(self, data):
-    self.week = sbp_messages.GPSTime(data).wn
-    self.nsec = sbp_messages.GPSTime(data).ns
+    self.week = MsgGPSTime(data).wn
```
mookerji: I don't have a great sense of the typing here: is `data` the payload of an SBP message or the entire SBP object? If it's only the payload, this won't work, because the constructor takes an entire SBP object.

mfine:
Oh, ping - this guy needs to come in, too. I'll rebase in a sec.

mfine:
Rebased.

scripts/baseline_view.py
mfine: The contract on per-message-type callbacks has been updated to take the entire message - updates have been made where appropriate to use `.payload` or to use the entire message, as here.

scripts/acq_results.py
```diff
@@ -67,10 +69,10 @@ def _receive_acq_result(self, data):
     self.acqs.append({})
     a = self.acqs[-1]
 
-    a['SNR'] = struct.unpack('f', data[0:4])[0]  # SNR of best point.
-    a['CP'] = struct.unpack('f', data[4:8])[0]   # Code phase of best point.
-    a['CF'] = struct.unpack('f', data[8:12])[0]  # Carr freq of best point.
-    a['PRN'] = struct.unpack('B', data[12])[0]   # PRN of acq.
+    a['SNR'] = struct.unpack('f', data.payload[0:4])[0]  # SNR of best point.
```
mookerji: See note above about `_receive_acq_result`.

scripts/baseline_view.py
```diff
@@ -24,7 +24,8 @@
 import datetime
 import time
 
-import sbp_piksi as sbp_messages
+from sbp.piksi      import SBP_MSG_IAR_STATE, SBP_MSG_INIT_BASE, SBP_MSG_RESET_FILTERS
```
mookerji: import here seems fine to me.

scripts/baseline_view.py
```diff
@@ -140,7 +141,7 @@ def _baseline_callback_ecef(self, data):
     return
 
   def iar_state_callback(self, data):
-    self.num_hyps = struct.unpack('<I', data)[0]
+    self.num_hyps = struct.unpack('<I', data.payload)[0]
```
mookerji: ... with updating the contract, this could be something like: 

```python
def iar_state_callback(self, sbp_msg):
  self.num_hyps = MsgIarState(sbp_msg).num_hyps
```

scripts/flash.py
```diff
@@ -11,13 +11,14 @@
...
 import serial_link
-import sbp_piksi as ids
 import struct
 import time
 import sys
-from itertools import groupby
 import new
 
+from itertools import groupby
+from sbp.piksi import SBP_MSG_FLASH_DONE, SBP_MSG_FLASH_ERASE, SBP_MSG_FLASH_READ, SBP_MSG_FLASH_PROGRAM, SBP_MSG_M25_FLASH_WRITE_STATUS, SBP_MSG_STM_FLASH_LOCK_SECTOR, SBP_MSG_STM_FLASH_UNLOCK_SECTOR
```
mookerji: Might want to break into separate lines.

scripts/solution_view.py
```diff
@@ -146,13 +146,14 @@ def update_table(self):
     self._table_list = self.table_spp.items()
 
   def pos_llh_callback(self, data):
-    soln = sbp_messages.PosLLH(data)
+    soln = MsgPosLLH(data)
```
mookerji: ... so the serial link callback here is using the full SBP message? If so, this will be clearer with a different name (like `sbp_msg` or something). 

Also, are you importing in the serial driver from libsbp?

scripts/acq_results.py
mfine: Yeah, wanted to leave this in initially before porting over to the new parsers. Believe there might be some other areas that could also benefit as well, but wanted a less risky iteration here.

scripts/baseline_view.py
mfine: Good point - will update to sbp_msg.

scripts/flash.py
mfine: Or I might add some wildcard imports.

scripts/solution_view.py
mfine: Its the old callback and its the full message. Will give it a better name.

mfine:
Going to bring this in to get things in-sync with master.

PR #274, opened 2015-03-18
Update exit code on serial_link KeyboardInterrupts.

mookerji:
We have a watchdog timer on Piksi heartbeats, but the
KeyboardInterrupts need return with `exit(1)` for us to see the error.

/cc @fnoble @mfine @cbeighley


mfine:
:shipit:

PR #276, opened 2015-03-19
Firmware update fixes

cbeighley:
Increase flashing speed (~5X on OSX 10.9)
Only update FPGA flash when there is new firmware
Only erase STM sectors once
Add pop-up progress bar to indicate flashing progress

### DIFF
--- a/scripts/flash.py
+++ b/scripts/flash.py
@@ -270,15 +270,15 @@ class Flash():
       for addr in reversed(range(start, end, ADDRS_PER_OP)):
         self.status = self.flash_type + " Flash: Programming address" + \
                                         " 0x%08X" % addr
-        if stream:
-          if mod_print == 0 or mod_print != 0 and self.print_count % mod_print == 0:
+        if mod_print == 0 or mod_print != 0 and self.print_count % mod_print == 0:
+          if stream:
             stream.write('\r' + self.status)
             stream.flush()
-            self.print_count = 1
-            if elapsed_ops_cb != None:
-              elapsed_ops_cb(self.ihx_elapsed_ops)
-          else:
-            self.print_count += 1
+          self.print_count = 1
+          if elapsed_ops_cb != None:
+            elapsed_ops_cb(self.ihx_elapsed_ops)
+        else:
+          self.print_count += 1
 
         binary = ihx.tobinstr(start=addr, size=ADDRS_PER_OP)
 
--- a/scripts/flash.py
+++ b/scripts/flash.py
@@ -17,8 +17,13 @@ import time
 import sys
 from itertools import groupby
 import new
+from intelhex import IntelHex
+from threading import Lock
 
+# 0.5 * ADDRS_PER_OP * MAX_QUEUED_OPS (plus SBP overhead) is how much each of
+# the STM32 TX/RX buffers will be filled by the program/read callback messages.
 ADDRS_PER_OP = 128
+MAX_QUEUED_OPS = 20
 
 M25_SR_SRWD = 1 << 7
 M25_SR_BP2  = 1 << 4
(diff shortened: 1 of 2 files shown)


cbeighley:
Have only tested this on OSX (10.9)

gsmcmullin:
Not working for me on Linux:

```
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
QPixmap: It is not safe to use pixmaps outside the GUI thread
[xcb] Unknown sequence number while processing reply
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
python: Fatal IO error 11 (Resource temporarily unavailable) on X server :0.
python: ../../src/xcb_io.c:635: _XReply: Assertion `!xcb_xlib_threads_sequence_lost' failed.
Aborted
```

scripts/update_view.py
```diff
@@ -99,6 +100,29 @@ def _choose_fw_fired(self):
     else:
       self.set_status('Error while selecting file')
 
+class PulsableProgressDialog(ProgressDialog):
+
+  def __init__(self, max, pulsed=False):
```
imh: nit: It's just style, but I don't like overriding builtin functions like max.

scripts/flash.py
```diff
@@ -100,41 +107,41 @@ def ihx_n_ops(ihx, addr_sector_map):
 
 # Lock sector of STM flash (0-11).
 def _stm_lock_sector(self, sector):
-  self._waiting_for_callback = True
   msg_buf = struct.pack("B", sector)
+  self.inc_n_queued_ops()
```
imh: Is there a way to make it less fragile than having to put inc_n_queued_ops and dec_n_queued_ops before/after the ops? Like making link.send_message do the work on n_queued_ops?

scripts/flash.py
```diff
@@ -154,114 +277,249 @@ def __init__(self, link, flash_type):
...
+    Get the current count of queued flash operation SBP messages.
+
+    Returns
+    -------
+    out : int
+      Get the current count of queued flash operation SBP messages.
+    """
+    return self._n_queued_ops
 
   def __del__(self):
```
fnoble: Its not recommended to put clean up in `__del__` in python. Consider using contexts instead:

http://stackoverflow.com/questions/865115/how-do-i-correctly-clean-up-a-python-object

fnoble:
Maybe out of scope for this PR but the protocol seems quite fragile. If we drop a `MSG_FLASH_DONE` packet then we have no way to know which operation it corresponded to. Also, right now if we drop a `MSG_FLASH_DONE` we will just wait forever for it to arrive.

scripts/flash.py
cbeighley: Will fix this.

cbeighley:
Agreed, it is fragile - and any way I can think of to make it less fragile would require a significant re-write of the module. Not a bad idea, but I think it's outside the scope of this PR as the fragility is basically the same as before (although if it failed after a read or program command you wouldn't know which it was ATM).

fnoble:
Chatted with Gareth about the fragility. I think its fine as we can always just start from the beginning and flash the firmware again on any kind of error.

See you changed to use `__enter__` etc. but that functionality is not being used currently. Need to wrap the usage of the class in a `with` block.

http://effbot.org/zone/python-with-statement.htm

Also, needs to be rebased onto master.

fnoble:
Other than that looks good to go.

cbeighley:
Current no piksi_firmware/scripts scripts use del on a flash object, so there's nothing to change - could change bootload.py to use `with` if you like.

PR #285, opened 2015-03-24
Update setup script to include libsbp.

mookerji:
/cc @fnoble @mfine


mfine:
Hmm, think we might publicly be on `0.24`:

https://pypi.python.org/pypi?%3Aaction=pkg_edit&name=sbp

think that's reflected in this https://github.com/swift-nav/libsbp/pull/39/files - which is what got pushed up - shameful (definitely not best practice :(

mookerji:
Well, I won't throw stones at your glass house if you don't throw them at mine.

PR #286, opened 2015-03-25
manage: Keep trying good satellites if no tracking channel is available.

gsmcmullin:
@fnoble hope this is in time for the code freeze.


src/manage.c
```diff
@@ -243,7 +243,8 @@ void manage_acq()
      * later using another fine acq.
      */
     log_info("No channels free :(\n");
-    acq_prn_param[prn].state = ACQ_PRN_TRIED;
+    acq_prn_param[prn].score = MIN(snr, 127);
```
fnoble: What if the SNR is really low, e.g. 20. This would still prioritize it over all other PRNs that haven't been tried?

PR #289, opened 2015-03-26
add argument to constructor for settings_list to specify location of set...

denniszollo:
...tings.yaml

Make settings file open call and path find  more robust


mfine:
Heads up! #354 is going to remove the `scripts/` directory and this commit contains `scripts/` changes. Before removing, let's get these changes in and brought over to `piksi_tools`.

PR #293, opened 2015-03-27
Use ephemeris from peer if ours is bad.

gsmcmullin:
This improves time to first fix, because we can use the ephemeris before it's received from the satellite.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -112,6 +112,26 @@ static msg_t nav_msg_thread(void *arg)
   return 0;
 }
 
+static void ephemeris_msg_callback(u16 sender_id, u8 len, u8 msg[], void* context)
+{
+  (void)sender_id; (void)context;
+
+  if (len != sizeof(ephemeris_t)) {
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -112,6 +112,26 @@ static msg_t nav_msg_thread(void *arg)
...
+{
+  (void)sender_id; (void)context;
+
+  if (len != sizeof(ephemeris_t)) {
+    log_warn("Received bad ephemeris from peer\n");
+    return;
+  }
+
+  gps_time_t t = get_current_time();
+  ephemeris_t e = *(ephemeris_t *)msg;
```
fnoble: should decode it as a `msg_ephemeris_t` from `libsbp/tracking.h` for future compatibility. Not sure about how to convert that into an `ephemeris_t`, perhaps add a function that just converts assuming the layout is the same but contains an `assert` checking the sizes are the same.

Also need to fix this where the ephemeris message is sent.

gsmcmullin:
Included in #309

PR #301, opened 2015-03-30
Acquire lock on flush of log file handle.

mookerji:
Your daily reminder that Python isn't thread-safe.

/cc @mfine @gsmcmullin


mfine:
Looks good.

PR #302, opened 2015-03-30
Drop unhealthy satellites.

gsmcmullin:


src/manage.c
```diff
@@ -284,6 +283,22 @@ u8 manage_track_new_acq(float snr)
   return MANAGE_NO_CHANNELS_FREE;
 }
 
+static void check_clear_unhealthy(void)
```
fnoble: This function needs a documentation comment.

src/manage.c
```diff
@@ -284,6 +283,22 @@ u8 manage_track_new_acq(float snr)
...
+{
+  static u16 wn;
+  gps_time_t t = get_current_time();
```
fnoble: Consider using `chTimeNow` instead as `get_current_time` reads from the FPGA

http://chibios.sourceforge.net/html/group__time.html

fnoble:
Looks good to me!

Issue #295, opened 2015-03-30
Move off serial_link.py


mfine:
Move scripts on to `libsbp` library and off of `serial_link.py`. Ephemeralize `serial_link.py`.

mfine:
Effort moved over to `piksi_tools` https://github.com/swift-nav/piksi_tools/pull/2.

PR #306, opened 2015-03-30
Lean out serial_link.py

mfine:
Remove `serial_link.py` in favor of sbp.client. This is a rough pass through weeding out the dependency on `serial_link.py` and moving the console out of piksi_firmware. There's definitely a lot more scope for refactoring here:

+ moving general purpose scripts into sbp.client (sbp.client.helpers?)
+ more cleanup and conventional patterns
+ use of message parsers from sbp.client.

This is towards fixing #295.

/cc @mookerji


mfine:
This is incomplete and not ready for review.

Issue #73, opened 2014-08-05
Global ephemeris struct is not thread safe


fnoble:
Currently we CPSID before writing but that doesn't help if we were in the process of reading and is quite heavy handed anyway.

Use a mutex to properly lock around reads and writes of the ephemeris struct.

soulcmdc:
@fnoble Can you confirm if #309 solved this issue?

PR #309, opened 2015-04-02
Ephemeris changes

gsmcmullin:
- Move ephemeris routines to a new file.
- Protect ephemeris structures with a mutex
- Keep confidence flags, set when the same ephemeris is decoded for a second time.
- Use peer's ephemeris if received over SBP.
- Revert to previous ephemeris if solution fails to converge.

### DIFF
--- a/src/ephemeris.c
+++ b/src/ephemeris.c
@@ -23,8 +23,33 @@
 
 MUTEX_DECL(es_mutex);
 ephemeris_t es[MAX_SATS] _CCM;
-static ephemeris_t es_old[MAX_SATS] _CCM;
-static bool es_confidence[MAX_SATS] _CCM;
+static ephemeris_t es_candidate[MAX_SATS] _CCM;
+
+static void ephemeris_new(ephemeris_t *e)
+{
+  gps_time_t t = get_current_time();
+  if (!ephemeris_good(&es[e->prn], t)) {
+    /* Our currently used ephemeris is bad, so we assume this is better. */
+    log_info("New untrusted ephemeris for PRN %02d\n", e->prn+1);
+    chMtxLock(&es_mutex);
+    es[e->prn] = es_candidate[e->prn] = *e;
+    chMtxUnlock();
+
+  } else if (memcmp(&es_candidate[e->prn], &e, sizeof(e))) {
+    /* The received ephemeris matches our candidate, so we trust it. */
+    log_info("New trusted ephemeris for PRN %02d\n", e->prn+1);
+    chMtxLock(&es_mutex);
+    es[e->prn] = *e;
+    chMtxUnlock();
+  } else {
+    /* This is our first reception of this new ephemeris, so treat it with
+     * suspicion and call it the new candidate. */
+    log_info("New ephemeris candidate for PRN %02d\n", e->prn+1);
+    chMtxLock(&es_mutex);
+    es_candidate[e->prn] = *e;
+    chMtxUnlock();
+  }
+}
 
 static WORKING_AREA_CCM(wa_nav_msg_thread, 3000);
 static msg_t nav_msg_thread(void *arg)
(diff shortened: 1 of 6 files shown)


gsmcmullin:
See @fnoble's comment in #293.

src/ephemeris.c
```diff
@@ -0,0 +1,164 @@
...
+
+        if (memcmp(&es[ch->prn], &e, sizeof(e))) {
+          if ((e.iode != es[ch->prn].iode) || /* IODE has changed, as expected */
+              !es_confidence[ch->prn]) {      /* Or we lack confidnce... */
+            log_info("New ephemeris for PRN %02d\n", ch->prn+1);
+            /* Back up old ephemeris in case this turns out to be bad. */
+            chMtxLock(&es_mutex);
+            /* Use the new ephemeris without confidence. */
+            es_old[ch->prn] = es[ch->prn];
+            es[ch->prn] = e;
```
imh: You don't have to memcpy here? Why is that?

imh:
lgtm

src/ephemeris.c
gsmcmullin: The compiler generates a call to memcpy to do this, but I think this is easier to read.

src/ephemeris.c
```diff
@@ -0,0 +1,164 @@
...
+          (ch->nav_msg.subframe_start_index == 0))
+        continue;
+
+      /* Decode ephemeris to temporary struct */
+      __asm__("CPSID i;");
+      s8 ret = process_subframe(&ch->nav_msg, &e);
+      __asm__("CPSIE i;");
+
+      if (ret < 0) {
+        log_info("PRN %02d ret %d\n", ch->prn+1, ret);
```
cbeighley: More descriptive log message here would be good - user won't know this has anything to do with ephemeris decoding at the moment.

cbeighley: Other than that lgtm

gsmcmullin: This wasn't changed from the old code, but I think this log message can be removed.  `process_subframe` complains with a log message about parity mismatch whenever it returns an error.

gsmcmullin:
These changes assume the bad ephemeris is so bad that the solution doesn't converge.  If it still converges we have no way to reject it.

src/ephemeris.c
cbeighley: Sounds good.

src/ephemeris.c
```diff
@@ -0,0 +1,162 @@
...
+
+  if (len != sizeof(ephemeris_t)) {
+    log_warn("Received bad ephemeris from peer\n");
+    return;
+  }
+
+  ephemeris_t e = *(ephemeris_t *)msg;
+  if (e.prn >= MAX_SATS) {
+    log_warn("Ignoring ephemeris for invalid sat\n");
+    return;
```
denniszollo: We should check that the simulator won't send out ephemeris.  If so, this might spam the console in simulation mode.

gsmcmullin:
The source of the bad ephemeris is likely to be an overrun of the subframe bits buffer.  See swift-nav/libswiftnav#114.

fnoble:
Hopefully this resolves #73

src/ephemeris.c
```diff
@@ -0,0 +1,162 @@
...
+          log_info("New ephemeris for PRN %02d\n", ch->prn+1);
+          /* Back up old ephemeris in case this turns out to be bad. */
+          chMtxLock(&es_mutex);
+          /* Use the new ephemeris without confidence. */
+          es_old[ch->prn] = es[ch->prn];
+          es[ch->prn] = e;
+          es_confidence[ch->prn] = false;
+          chMtxUnlock();
+        } else {
+          log_info("Ignoring new ephemeris for PRN %02d "
```
fnoble: Suggest making this a `log_warn`

fnoble: Then we can track it more easily in HITL

src/ephemeris.c
```diff
@@ -0,0 +1,162 @@
...
+
+  static sbp_msg_callbacks_node_t ephemeris_msg_node;
+  sbp_register_cbk(
+    SBP_MSG_EPHEMERIS,
+    &ephemeris_msg_callback,
+    &ephemeris_msg_node
+  );
+
+  chThdCreateStatic(wa_nav_msg_thread, sizeof(wa_nav_msg_thread),
+                    NORMALPRIO-1, nav_msg_thread, NULL);
```
fnoble: Check this thread is higher priority than `time_matched_obs` thread to help ensure it meets its timing deadline.

src/ephemeris.c
fnoble: Sim doesn't send ephemerides so this should be fine

src/ephemeris.c
```diff
@@ -0,0 +1,162 @@
...
+      __asm__("CPSID i;");
+      s8 ret = process_subframe(&ch->nav_msg, &e);
+      __asm__("CPSIE i;");
+
+      if (ret <= 0)
+        continue;
+
+      /* Decoded a new ephemeris. */
+      if (memcmp(&es[ch->prn], &e, sizeof(e))) {
+        if ((e.iode != es[ch->prn].iode) || /* IODE has changed, as expected */
```
fnoble: If IODE has changed we could continue to use the previous ephemeris until we have confidence in the new version. This would remove the remaining vulnerability to when the ephemeris IODE is corrupted.

Old ephemeris is still valid for a while even after a new one is transmitted.

fnoble:
Need to update `base_obs.c` which also uses the ephemerides.

https://github.com/gsmcmullin/piksi_firmware/blob/ephemeris/src/base_obs.c#L281
https://github.com/gsmcmullin/piksi_firmware/blob/ephemeris/src/base_obs.c#L294

Also would be good to double check they aren't used anywhere else (don't think they are).

src/ephemeris.c
gsmcmullin: It is.  No change needed.

src/ephemeris.c
gsmcmullin: The latest commit added changes this logic.  IODE is not checked explicitly, but the entire ephemeris struct is saved as a candidate until it's received a second time.  The old ephemeris is no longer kept around, and the logic to revert has been removed.

fnoble:
This is ready to go! Just need to coordinate with @cbeighley on merging.

fnoble:
Updated changes look good. Just did a bench test and seems to be working.

PR #312, opened 2015-04-02
[PYINSTALLER] Changes to console.spec to accomodate SBP and settings changes

cbeighley:
Handle libsbp and settings updates


mfine:
Looks good to me - as long as the analysis doesn't need to see any of the stuff that was in `../`

PR #321, opened 2015-04-07
Update sys_deps.yml fixing 'construct' dependency error

### LINKED ISSUE #320: setup.sh (via ansible playbook) fails to install 'construct', required by 'sbp'
Attempting to use 'setup.sh -x install' on a fresh machine (tested: Ubuntu 12.04, Debian Jessie) fails at 

TASK: [common_ubuntu | Install python Enthought dependencies]

[Full log](https://gist.github.com/henryhallam/5c637b3c0e43545dae29)

The problem appears to be that the '[construct](https://pypi.python.org/pypi/construct)' package is needed by '[sbp](https://pypi.python.org/pypi/sbp/0.27)''s setup.py - it is not merely a runtime dependency of the package.  In Debian parlance it might be considered a 'Pre-Depends'.  The ansible playbook lists both 'construct' and 'sbp' in the same invocation of 'pip install' (setup/ansible/roles/common_ubuntu/tasks/sys_deps.yml line 55):
```
pip install traits traitsui pyserial pylibftdi
             pyparsing==1.5.7 pygments intelhex six construct sbp==0.27
```

pip 'collects' (i.e. downloads) the tarball for each package, in the order given, but does not attempt to install them in the same order.

As a workaround, running 'sudo pip install construct' and rerunning setup.sh will do the job.

henryhallam:
Fixes #320 

@mookerji @cbeighley for review

### DIFF
--- a/setup/ansible/roles/common_ubuntu/tasks/sys_deps.yml
+++ b/setup/ansible/roles/common_ubuntu/tasks/sys_deps.yml
@@ -48,14 +48,11 @@
   sudo: Yes
   apt: update_cache=yes
 
-- name: Install python Enthought dependencies
+- name: Install python dependencies
   sudo: True
   command: >
              pip install traits traitsui pyserial pylibftdi
              pyparsing==1.5.7 pygments intelhex six construct sbp==0.27
-             --allow-unverified PIL
-             --allow-external intelhex
-             --allow-unverified intelhex
 
 # Set up conditional installation for ARM GCC tooling later.
 - name: Checking for arm gcc


setup/ansible/roles/common_ubuntu/tasks/sys_deps.yml
```diff
@@ -48,14 +48,15 @@
   sudo: Yes
   apt: update_cache=yes
 
-- name: Install python Enthought dependencies
+- name: Install python dependencies
   sudo: True
   command: >
```
henryhallam: Should we be using the 'apt' ansible module for this, rather than the generic 'command' module?

setup/ansible/roles/common_ubuntu/tasks/sys_deps.yml
```diff
@@ -48,14 +48,15 @@
...
              pip install traits traitsui pyserial pylibftdi
-             pyparsing==1.5.7 pygments intelhex six construct sbp==0.27
```
mookerji: ... so it does bother me that the `--allow-unverified` is needed here, I don't think that intelhex will install otherwise. Is that no longer the case?

mfine:
`sbp==0.29` should remove the need for this change.

setup/ansible/roles/common_ubuntu/tasks/sys_deps.yml
henryhallam: I confirmed that all packages install OK without the --allow-unverified etc.  Presumably upstream fixed that at some point.  I can reconfirm on a fresh machine if you like.

henryhallam:
@mfine @mookerji OK, I've undone the installation of SBP as a separate step.  Retained the removal of unnecessary/unsafe pip install flags.  I do still wonder if it would be better to somehow separate "our" packages (i.e. sbp) from other dependencies.

mfine:
@henryhallam having our deps in a separate step sounds fine to me.

setup/ansible/roles/common_ubuntu/tasks/sys_deps.yml
mookerji: Yeah...

mookerji:
This looks good to me. I do agree that perhaps having the swiftnav dependencies in a separate role is a good idea. We can probably do that after the piksi console repository is up and running.

mookerji:
retest this please

PR #334, opened 2015-04-09
External event capture for photogrammetry etc.

henryhallam:
There's a new setting, ext_events.edge_trigger, that selects which (if any)
edges on DEBUG0 will trigger a timestamped event report.

Presently, events are output as an INFO message.  A new SBP message type is
being defined to replace that.


henryhallam:
Code review: @fnoble @cbeighley @soulcmdc @gsmcmullin 

Todo:
* Waiting on SBP stability to use MSG_EXT_EVENT rather than an info message
* Convert to GPS time.

henryhallam:
This is working nicely, I'm happy with the structure and it's ready for review. @fnoble @cbeighley @denniszollo

mfine:
Heads up! #354 is going to remove the `scripts/` directory and this commit contains `scripts/` changes. Before removing, let's get these changes in and brought over to `piksi_tools`.

src/board/nap/nap_common.h
```diff
@@ -66,6 +66,17 @@ void nap_conf_b_clear(void);
 void nap_xfer_blocking(u8 reg_id, u16 n_bytes, u8 data_in[],
                        const u8 data_out[]);
 
+/** Convenience function to read 4 bytes from a register (writing zeros) and
+ * convert to host byte order (i.e. little-endian).
+ * \param reg_id   NAP register ID.
+ * \return u32 register value
+ */
+inline u32 nap_read_u32(u8 reg_id) {
```
fnoble: `{` on its own line

src/board/nap/nap_ext_event.c
```diff
@@ -0,0 +1,97 @@
...
+} trigger_type_t;
+static trigger_type_t trigger = NONE;
+
+static bool trigger_changed(struct setting *s, const char *val)
+{
+  if (s->type->from_string(s->type->priv, s->addr, s->len, val))
+  {
+    u8 v[5]={0};
+    v[4] = trigger;
+    nap_xfer_blocking(NAP_REG_EXT_EVENT_TIME, 5, v, v);
```
fnoble: Maybe make a function to set this register and move to `nap_common`

src/board/nap/nap_ext_event.c
```diff
@@ -0,0 +1,97 @@
...
+#include "../../settings.h"
+#include "../../timing.h"
+#include "../../sbp.h"
+
+/** \defgroup ext_events External Events
+ * Capture accurate timestamps of external pin events
+ * \{ */
+
+typedef enum {
+  NONE,
```
fnoble: If you rely on the numerical value of these then you should explicitly define them

src/board/nap/nap_ext_event.c
```diff
@@ -0,0 +1,97 @@
...
+void ext_event_service(void)
+{
+  union {
+    struct __attribute__((packed)) {
+      u8 edge_pin;
+      u32 time;
+    } d;
+    u8 b[5];
+  } v;
+  v.b[4] = trigger;  // We have to reset the trigger for next time
```
fnoble: change all `//` to `/* */`

src/board/nap/nap_ext_event.c
```diff
@@ -0,0 +1,97 @@
...
+  // We have to infer the most sig word (i.e. # of 262-second rollovers)
+  u64 tc = nap_timing_count();
+  if ((tc & 0xFFFFFFFF) < event_nap_time) // Rollover occurred since event
+    tc -= UINT64_C(0x100000000);
+  event_nap_time |= tc & UINT64_C(0xFFFFFFFF00000000);
+
+  gps_time_t gpst = rx2gpstime(event_nap_time);
+  msg_ext_event_t msg;
+  msg.wn = gpst.wn;
+  msg.tow = gpst.tow * 1E3;
```
fnoble: pedantry: we tend to use `1e3`

fnoble:
Looks good. One thought - we were trying to keep everything under `board/nap` as more of a library for interfacing with the NAP.

I would suggest making a function to read/write `NAP_REG_EXT_EVENT_TIME` in `nap_common.c` and then you can rename `src/board/nap/nap_ext_event.c` to `src/ext_event.c`. This keeps the firmware/OS code separate from the low level NAP read/write code.

If you wanted to split the `ext_event` functions out from `nap_common` then you could have a file called `src/board/nap/nap_ext_event.c` but I think it should just contain the register interface code.

henryhallam:
@fnoble OK, moved into main directory and rebased on master.  Bench testing looks fine, still showing the same occasional glitches.  I'll investigate those tomorrow but I think they're unrelated to this functionality and symptomatic of something in the PVT solver or elsewhere.

Issue #353, opened 2015-04-13
Fix half cycle slips


cbeighley:
Check DD observations to verify half cycle and discontinuity bugs are fixed

soulcmdc:
@fnoble was this fixed by #356 ?

fnoble:
@soulcmdc need to verify, should have been fixed by #413

PR #356, opened 2015-04-14
Reset ambiguity when SNR falls below threshold

fnoble:
Hopefully fixes issues with frequent half cycle slips seen in double difference observations when SNR falls below the threshold and then picks up again.

cc @gsmcmullin @henryhallam @imh

### DIFF
--- a/src/track.c
+++ b/src/track.c
@@ -295,12 +295,29 @@ void tracking_channel_update(u8 channel)
  * Change tracking channel state to TRACKING_DISABLED and write 0 to SwiftNAP
  * tracking channel code / carrier frequencies to stop channel from raising
  * interrupts.
+ *
  * \param channel Tracking channel to disable.
  */
 void tracking_channel_disable(u8 channel)
 {
   nap_track_update_wr_blocking(channel, 0, 0, 0, 0);
   tracking_channel[channel].state = TRACKING_DISABLED;
+}
+
+/** Sets a channel's carrier phase ambiguity to unknown.
+ * Changes the lock counter to indicate to the consumer of the tracking channel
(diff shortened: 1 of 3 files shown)


fnoble:
Fixes #353 (hopefully!)

src/track.c
```diff
@@ -303,6 +304,22 @@ void tracking_channel_disable(u8 channel)
...
+ * invalidates the time of week to indicate that the half cycle ambiguity must
+ * be resolved again by the navigation message processing. Should be called if
+ * a cycle slip is suspected.
+ *
+ * \param channel Tracking channel to disable.
+ */
+void tracking_channel_ambiguity_unknown(u8 channel)
+{
+  u8 prn = tracking_channel[channel].prn;
+  tracking_channel[channel].TOW_ms = -1;
```
henryhallam: Should we have a
```
#define TOW_INVALID -1
```
?

fnoble:
Bench tested, no obvious regressions. Not verified if it reduces the half cycle slip rate but @mookerji was planning to do the analysis when we have accumulated some data on HITL.

@henryhallam any further comments or do you think this is ok to merge?

henryhallam:
track.c:tracking_channel_init():
```
  tracking_channel[channel].lock_counter = ++tracking_lock_counters[prn];

  /* Use -1 to indicate an uninitialised value. */
  tracking_channel[channel].TOW_ms = -1;
```
May as well replace those lines with a call to tracking_channel_ambiguity_unknown().

Also, and this is actually a (minor) bug: manage.c:use_tracking_channel() should test TOW_ms against TOW_INVALID rather than against 0.

henryhallam:
Aside from that, I say go for it.

PR #361, opened 2015-04-14
Write JSON logs in addition to pickled logs

mfine:
Rearrange the logic so that JSON logs are configured independent of pickled logs. This lets us generate both logs while we're transitioning from JSON logs to pickled logs. The goal is to get the `-j` flag turned on for analysis without breaking existing analysis.

Part of https://github.com/swift-nav/sbp_log_analysis/issues/97.

/cc @mookerji


mfine:
Going to bring this in. Not much going on here - just writing out JSON logs in addition to pickled logs.

Issue #328, opened 2015-04-07
Improve hard fault handlers


fnoble:
Right now I think they just busy wait unless the debugger is attached. They should print an error and reboot.

Issue #327, opened 2015-04-07
Handle calls to abort


fnoble:
Right now we just busy wait (the default newlib implementation), we should print an error and reset the device.

Issue #330, opened 2015-04-07
Screaming death should reset the device


fnoble:

henryhallam:
Let's just let the WDT reset the device.  That way, resets can be disabled by disabling the WDT setting if a developer wants to debug.  It also gives more of a chance for the death message to be transmitted over any radio links involved.

Issue #329, opened 2015-04-07
Implement watchdog timer


henryhallam:
Use Window Watchdog.  In EWI, print error message before hardware reboots.

Open: which thread should clear watchdog?

henryhallam:
/cc @fnoble @gsmcmullin Comments requested before I get too far with implementation...

Rationale: As part of general fault handling cleanup, we want to (1) know when and why a crash occurred, and (2) not leave the system permanently in a crashed state.  Using the hardware watchdog timer is the most foolproof way of accomplishing the latter, but there are some subtleties.

The simplest option is to clear the WDT in e.g. the main loop or the heartbeat (system monitor) thread.  That will catch certain kinds of lockup but not others, where one of the other threads might spin forever but Chibi will preempt it and allow the heartbeat thread to keep clearing the WDT.  So instead we want the WDT-clearing thread to require some kind of affirmative signal from each system thread, and check that all those signals are present before clearing the WDT.

My current plan is to use ChibiOS Events, in a similar manner to that [discussed here](http://forum.chibios.org/phpbb/viewtopic.php?f=2&t=2394), where they are applied to exactly the same situation.  The heartbeat thread loop will be the only place that the hardware WDT is cleared, and that loop will chEvtWaitAllTimeout() on the flags from each thread before doing so.   Under normal conditions that syscall will return immediately, because the heartbeat loop period is longer than the period of all the events we want to watch over.  If any of the other threads are locked up, it will return after a timeout, which we can detect and use to report which thread is the problem.

The STM32F40x has two watchdog timer peripherals, the Independent Watchdog (IWDG) and the Window Watchdog (WWDG).  The WWDG has the rather attractive feature of optionally triggering an interrupt shortly before resetting the chip upon timer expiry, which could be used to report that a watchdog reset is about to occur (and possibly what caused it).  Unfortunately there are two drawbacks to the WWDG: 1. It runs from the APB1 clock, which is ultimately derived from the front end (via the FPGA).  A fault in the front end configuration or in the FPGA could cause this to stop, and the watchdog would never be triggered.  2. It has a max period of 64 ms.  Events we'd like to watch over may occur less frequently than that (e.g. nav solution epochs), thus we'd need complicated logic in the watchdog thread which would somewhat defeat the objective.

[...]

PR #369, opened 2015-04-16
Fault-handling (including watchdog)

henryhallam:
This pull req will ultimately address #326 #327 #328 #329 #330.

### DIFF
--- a/src/peripherals/usart.c
+++ b/src/peripherals/usart.c
@@ -185,16 +185,16 @@ void usarts_enable(u32 ftdi_baud, u32 uarta_baud, u32 uartb_baud, bool do_precon
   if (do_preconfigure_hooks) {
 
     /* TODO: Should this really be here? */
-    log_info("Piksi Starting...\n");
     if ((RCC_CSR & 0xFF000000) != RCC_CSR_PINRSTF) {
-      log_info("Reset reason: %02X\n", (unsigned int)(RCC_CSR >> 24));
       if (RCC_CSR & RCC_CSR_IWDGRSTF)
-        log_warn("Watchdog reset detected.\n");
+        log_error("Piksi has reset due to a watchdog timeout.\n");
       if (RCC_CSR & RCC_CSR_LPWRRSTF)
-        log_warn("Low power reset detected.\n");
+        log_error("Low power reset detected.\n");
       if (RCC_CSR & RCC_CSR_SFTRSTF)
         log_info("Software reset detected.\n");
+      log_info("Reset reason: %02X\n", (unsigned int)(RCC_CSR >> 24));
     }
+    log_info("Piksi Starting...\n");
     RCC_CSR |= RCC_CSR_RMVF;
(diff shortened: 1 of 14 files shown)


henryhallam:
Screaming death: Closes #326 
_exit handler: Closes #327

henryhallam:
Fault handler: Closes #328 
Would be nice to add more trace functionality to this, but it should already tell us much more than "hurr durr system crashed"

@fnoble @gsmcmullin review plz

henryhallam:
@fnoble I'm not going to add fallback_write for all the UARTs right now, but I've put a TODO for that as well as the more extensive stack/state dump.  The only hazard I can think of with this is that the integer-divide-by-zero trapping is new, and perhaps there are some divide-by-zeros occurring occasionally that we didn't notice before, which will now result in a crash.  I think we'd want to know about them, though.  FWIW in single-ended mode with a roof antenna, that is not occurring.

So, OK to merge?

fnoble:
Looks good, just going to run it on HITL for a while

src/peripherals/usart.c
```diff
@@ -184,7 +184,18 @@ void usarts_enable(u32 ftdi_baud, u32 uarta_baud, u32 uartb_baud, bool do_precon
 
   if (do_preconfigure_hooks) {
 
+    /* TODO: Should this really be here? */
     log_info("Piksi Starting...\n");
+    if ((RCC_CSR & 0xFF000000) != RCC_CSR_PINRSTF) {
+      log_info("Reset reason: %02X\n", (unsigned int)(RCC_CSR >> 24));
+      if (RCC_CSR & RCC_CSR_IWDGRSTF)
+        log_warn("Watchdog reset detected.\n");
```
fnoble: Make this a `log_error` as discussed

src/peripherals/usart.c
```diff
@@ -184,7 +184,18 @@ void usarts_enable(u32 ftdi_baud, u32 uarta_baud, u32 uartb_baud, bool do_precon
...
+      if (RCC_CSR & RCC_CSR_LPWRRSTF)
+        log_warn("Low power reset detected.\n");
```
fnoble: I'd think about making this an error too?

fnoble:
@henryhallam : Could we try running this on HITL with the timeout set much longer (e.g. 20s)? Then we can catch the irrecoverable crashes for now and get this in master before the locking issues are fixed (which might take some time).

Also would appreciate if you could rebase and make those last changes to the `log_warn`s. Then I think this is good to go!

henryhallam:
Done.  WDT hardware period can't be > 32768 ms so I made it 30 seconds and the thread period 15 seconds.  Do you want to run it on the mega HITL or just the regular automatic PR HITL?

fnoble:
@henryhallam : Awesome, I'll set it going on mega HITL now

fnoble:
HITL looks good

PR #377, opened 2015-04-20
More consistent heartbeat period

fnoble:
Previously we were waiting 1s since the task completed, now we keep track of the previous time we woke up and sleep until one second since that point.

cc: @henryhallam @gsmcmullin @denniszollo

### DIFF
--- a/src/system_monitor.c
+++ b/src/system_monitor.c
@@ -100,14 +100,41 @@ static msg_t track_status_thread(void *arg)
   return 0;
 }
 
+/** Sleep thread until a period has elapsed.
+ * Keeps track of the previous wake-up time to ensure that a periodic task is
+ * woken up on time even if there is jitter in the time the task takes to
+ * execute (e.g. due to preemption by higher priority tasks).
+ *
+ * References:
+ *  -# https://www.rfc1149.net/blog/2013/04/03/sleeping-just-the-right-amount-of-time/
+ *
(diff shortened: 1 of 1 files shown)


src/system_monitor.c
```diff
@@ -100,14 +100,41 @@ static msg_t track_status_thread(void *arg)
   return 0;
 }
 
+/** Sleep thread until a period has elapsed.
+ * Keeps track of the previous wake-up time to ensure that a periodic task is
+ * woken up on time even if there is jitter in the time the task takes to
+ * execute (e.g. due to preemption by heigher priority tasks).
```
henryhallam: spelling s/heigher/higher/

src/system_monitor.c
```diff
@@ -100,14 +100,41 @@ static msg_t track_status_thread(void *arg)
   return 0;
 }
 
+/** Sleep thread until a period has elapsed.
+ * Keeps track of the previous wake-up time to ensure that a periodic task is
+ * woken up on time even if there is jitter in the time the task takes to
+ * execute (e.g. due to preemption by higher priority tasks).
+ *
+ * References:
+ *  -# https://www.rfc1149.net/blog/2013/04/03/sleeping-just-the-right-amount-of-time/
```
henryhallam: yay reference

henryhallam:
Looks good.

PR #380, opened 2015-04-25
Update to latest libswiftnav

fnoble:
Includes:
 - More conservative altitude limit
 - Removed LAPACKe
 - Clean up of static qualifiers and some dead code
 - Plover basic support
 - Hooks for inspecting filter state from python
 - Support for 20ms integration in new nag msg sync


fnoble:
HITL testing looks good

PR #390, opened 2015-04-29
Support masking satellites for acquisition and tracking.

gsmcmullin:
This adds support for a new SBP message MSG_MASK_SATELLITE which allows a mask to be set preventing a satellite from being acquired, and forcing a tracked satellite to be dropped.

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -16,6 +16,7 @@
 #include <ch.h>
 
 #include <libsbp/sbp.h>
+#include <libsbp/piksi.h>
 #include <libswiftnav/logging.h>
 #include <libswiftnav/almanac.h>
 #include <libswiftnav/constants.h>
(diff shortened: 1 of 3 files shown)


cbeighley:
retest this please

cbeighley:
Looks good to me. One thought I had is it might be better to mess up the correlations for the PRN and let the loop filters change the frequencies as a result of that, but it would require more state. @fnoble @henryhallam opinions?

fnoble:
lgtm. I think changing the frequency is good, with the correlation you would have to keep track of the mask state in order to continue to mess it up for a period of time and it would get more convoluted.

henryhallam:
I'm a little confused as to why we don't just directly drop the satellite by calling tracking_channel_disable()?

cbeighley:
I think the idea is to exercise the entire chain of logic on the device and have the receiver decide to drop the satellite, rather than bypassing that logic and deciding for it.

henryhallam:
That does sound worth having, but to me that's a different thing from masking it.  I would implement that as a "nuke channel" *message*, and have the "mask satellite" *setting* simply disable the channel if it's running, and prevent it from being searched during acq.

gsmcmullin:
Replaced by #462

PR #391, opened 2015-05-02
Update to latest libswiftnav

cbeighley:


cbeighley:
Pulls in Scott's libswiftnav updated sat inclusion process to use s64's to avoid overflow issue

Issue #372, opened 2015-04-16
Default antenna selection to "External"


denniszollo:
We don't want any users to use the patch anyway, so why default to Auto?

PR #395, opened 2015-05-06
Set default antenna selection to External

denniszollo:
Customers have reported issues where the MAX's AUTO selection sometimes fails.  

The patch antenna shouldn't be used, so we default to EXTERNAL.

This should address: #372


fnoble:
lgtm

PR #405, opened 2015-05-15
Bring into line with libswiftnav dgnss state refactor

henryhallam:
Untested!

It seems a little disingenuous to initalize the whole dgnss_state only on an as-needed basis when we gain > 4 sats.  I'd be inclined to do that in solution_setup() instead.  Thoughts?

/cc @fnoble @mookerji

### DIFF
--- a/src/STM32F405xG.ld
+++ b/src/STM32F405xG.ld
@@ -122,7 +122,6 @@ SECTIONS
 
     .ccmram (NOLOAD) : {
       *(.ccmram*)
-      *(.bss.nkf)
     } > ccmram
 
     .data :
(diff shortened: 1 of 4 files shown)


mookerji:
I agree. I think the dgnss initialization should happen with a canonical setup.

henryhallam:
retest this please

fnoble:
lgtm (after HITL testing)

denniszollo:
retest this please

henryhallam:
Appears to cause hard faults on Piski 1227 in the test rack, but not on Piksi 1757 on my desk...

PR #407, opened 2015-05-18
Use high elevation sat for nominal range

cbeighley:
Michael Oborne's libswiftnav fix: Use high elevation sat for nominal range


cbeighley:
Closed in lieu of #407

Issue #81, opened 2014-08-18
Still dropping out of Fixed mode when resolving occasionally


fnoble:
Sometimes even when not gaining a satellite.

imh:
1) is getting log files containing the problem
2) run through SITL

mookerji:
I'm keeping a journal of this issue, which I'm going to link here later today. The current plan is: 

0. Find a reproducible test case for using external inputs, since it seems like a control flow problem.
1. Add a HITL test for inappropriate fixed mode transitions. Collect overnight instances of logs where that happens. Might want to see how many fixed=>float transitions result from this, versus other situations (like loss of tracking).
2. Do some trace debugging for the `dgnss_update` methods to eliminate the current hypothesis.
3. SITL checking of the filter with the existing log.
4. Add some mocked filter testing to the `dgnss_update` based on results of 2 and 3.

fnoble:
I think (5) should probably be in the form of a harness to automatically run a set of log files through SITL and testing of the results. Lets chat about it when you have a free moment.

fnoble:
Current status: fixed -> float transitions have been reduced a lot but not entirely eliminated, the remainder are mostly due to false locks and satellite signal loss (rather than bugs) but this needs to be properly confirmed

mookerji:
What fergus said. I think the remaining cases would need to be categorized.

PR #408, opened 2015-05-21
Observations output below CN0 threshold

fnoble:
**Why:** We have observed observations being generated with CN0 values below the tracking threshold specified in the firmware. These observations also frequently contain cycle slips (as may be expected given the low SNR). It is hypothesized that these observations are one of the main causes of fixed -> float transitions.

**How:** Right now we don't directly test the SNR when checking if we should filter out an observation, we only check the hysteresis counter. I think that what we are seeing is somehow the `manage_track()` thread is not getting called in time to update and hence we are outputting bad measurements. The easiest fix is to just directly check the SNR each time in addition to looking at the hysteresis counter.

**Testing:** HITL test has been created to check for this effect, see `TrackingTest_observations_obs_cn0_below_thres` and `TrackingTest_observations_obs_cn0_min`.

Hopefully helps with #81.

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -448,12 +448,21 @@ static void manage_track()
 s8 use_tracking_channel(u8 i)
 {
   return (tracking_channel[i].state == TRACKING_RUNNING)
+      /* Check ephemeris is useable. */
+      /* TODO: Use libswiftnav function here. */
       && (es[tracking_channel[i].prn].valid == 1)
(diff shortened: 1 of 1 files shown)


fnoble:
HITL indicates a large (but not total) drop in fixed->float transitions.

I'm happy to put this up for review now.

@gsmcmullin @imh @mookerji

imh:
lgtm

PR #412, opened 2015-05-29
update libswiftnav pointer to integrity check testing branch

kovach:
for HITL testing

libswiftnav code unfinished


gsmcmullin:
This looks like an abandoned test head.  @kovach reopen if there was something important here.

PR #413, opened 2015-05-30
Wait for 5s after mode changes for transients to stabilize

fnoble:
**Why:** We have observed strange things happening when satellites first start outputting observations. Often we will see a half-cycle slip within a few data points of the beginning and sometimes we will get only a few data points for a satellite that appears not to have phase lock. These effects are causing serious mischief in the filters.

It is hypothesized that this is because we transition to 20ms tracking right at the instant we start outputting observations.

**How:** Wait for a while (5s) after "mode changes" to allow the tracking loops time to stabilize and ride over any strange transients that occur at these events. This also gives time for the alias detection to do its job.

Mode changes right now are:
 - Changes to integration length
 - Frequency adjustments for alias lock correction

This is implemented by adding a field `mode_change_count` to the state struct that is set equal to `update_count` on any events expected to cause transients. In `use_tracking_channel()` we now compute the different between `mode_change_count` and `update_count` to check sufficient time has elapsed since the mode change.

**Testing:** HITL

cc: @gsmcmullin @imh @mookerji @cbeighley

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -456,6 +456,11 @@ s8 use_tracking_channel(u8 i)
       && (tracking_channel[i].update_count
             - tracking_channel[i].snr_below_threshold_count
             > TRACK_SNR_THRES_COUNT)
+      /* Check that a minimum time has elapsed since the last tracking channel
+       * mode change, to allow any transients to stabilize. */
+      && (tracking_channel[i].update_count
+            - tracking_channel[i].mode_change_count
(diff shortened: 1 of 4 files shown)


fnoble:
HITL looks good, @cbeighley want to take a look to?

@gsmcmullin : ready for review when you have time!

gsmcmullin:
lgtm

cbeighley:
lgtm

PR #418, opened 2015-06-03
Improve tracking over mechanical stress.

gsmcmullin:
This uses a faster responding C/N0 estimator and changes to the alias detect parameters to improve tracking over mechanical stress.  This is showing fairly good results for me at 10ms integration.  The integration time can be lowered for better performance.


cbeighley:
Merged via #440

PR #422, opened 2015-06-07
Print thread states on watchdog thread timeout

fnoble:
cc: @henryhallam 

Tested 2 ways:

1) Removed the call to `watchdog_notify()` in `manage_track_thread`
2) Removed the call to `chSemSignal()` in `cq_service_irq()` and increased the timeouts.

Both worked as expected. In case (2) the address of `acq_pipeline_sem` was in `wtobjp` as expected.

### DIFF
--- a/src/system_monitor.c
+++ b/src/system_monitor.c
@@ -171,6 +171,34 @@ static msg_t system_monitor_thread(void *arg)
   return 0;
 }
 
+static void debug_threads()
+{
+  const char* state[] = {
+    "READY",
+    "CURRENT",
+    "SUSPENDED",
+    "WTSEM",
+    "WTMTX",
+    "WTCOND",
+    "SLEEPING",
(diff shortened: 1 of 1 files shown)


src/system_monitor.c
```diff
@@ -171,6 +171,34 @@ static msg_t system_monitor_thread(void *arg)
...
+    "WTANDEVT",
+    "SNDMSGQ",
+    "SNDMSG",
+    "WTMSG",
+    "WTQUEUE",
+    "FINAL"
+  };
+  Thread *tp = chRegFirstThread();
+  while (tp) {
+  log_info("%s (%u: %s): prio: %lu, flags: %u, wtobjp: 0x%08X\n",
```
henryhallam: 'wtobjp: %p' saves four bytes of flash, and you don't need the cast to size_t

PR #427, opened 2015-06-08
Baseline Refactor

### LINKED ISSUE #141: Solution thread is missing its timing deadline
The `time_matched_obs` thread is currently holding the `base_obss_lock` for a long time, causing the solution thread to very occasionally miss its timing deadline.

Moving `process_matched_obs()` out of the lock zone improves system liveness considerably and fixes an issue where the solution thread will miss its timing deadline. However, it would introduce an (unlikely) possibility that the low latency baseline could be calculated when the filter is in the process of being updated.

The proposed fix would be to have the low latency baseline calculation use a local copy of the ambiguity state that is updated with a much cheaper call that just copies the state out of the filter. This update would be called every time directly after `dgnss_update()`. The lock zone could then be just around this update call.

fnoble:
Fixes #141

Pulls in swift-nav/libswiftnav#149

### DIFF
--- a/src/nmea.c
+++ b/src/nmea.c
@@ -93,8 +93,8 @@ u8 nmea_checksum(char *s)
  * \param soln Pointer to gnss_solution struct.
  * \param dops Pointer to dops_t struct.
  */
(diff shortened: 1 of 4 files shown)


fnoble:
@cbeighley : this is ready for HITL testing

PR #428, opened 2015-06-09
Update libswiftnav to swiftnav/master with new initial covariance.

cbeighley:


cbeighley:
Testing looks good, albeit little difference.

PR #431, opened 2015-06-10
Michael Oborne's high elevation sat fix

cbeighley:


cbeighley:
Already merged via updating libswiftnav in #428

Issue #432, opened 2015-06-10
Remove flash callbacks


cbeighley:
You should only be able to modify the flash directly while the device is in the bootloader.

PR #433, opened 2015-06-10
Remove flash

cbeighley:
Remove flash callbacks from application firmware. You should only be able to modify the flash directly when the device is in the bootloader.

Maybe in libsbp the flash callbacks should be moved under the bootloader?

@fnoble @mfine


cbeighley:
I've created an issue for moving the flash messages under the bootloader section in libsbp: https://github.com/swift-nav/libsbp/issues/164.

Closes #432

src/peripherals/stm_flash.c
```diff
@@ -117,6 +121,26 @@ u8 stm_flash_program(u32 address, u8 data[], u8 length)
   return FLASH_OK;
 }
 
+/** Callback to read STM32F4's hardcoded unique ID.
+ * Sends STM32F4 unique ID (12 bytes) back to host.
+ */
+void stm_unique_id_callback(u16 sender_id, u8 len, u8 msg[], void* context)
```
fnoble: I don't think this should be in here? Perhaps just move to its own file or to `init.c`?

mfine: This callback is going to continue to be done both in the bootloader and the firmware?

src/peripherals/stm_flash.c
```diff
@@ -12,10 +12,14 @@
 
 #include <libopencm3/stm32/f4/flash.h>
 
+#include <libsbp/flash.h>
+
 #include "stm_flash.h"
 #include "../error.h"
-#include "../flash_callbacks.h"
 #include "main.h"
+#include "../sbp.h"
```
fnoble: This file shouldn't need to use SBP I think

src/flash.h
```diff
@@ -10,11 +10,8 @@
  * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
  */
 
-#ifndef SWIFTNAV_FLASH_CALLBACKS_H
-#define SWIFTNAV_FLASH_CALLBACKS_H
-
-#define FLASH_STM 0 /**< Value to pass flash callbacks to use STM Flash */
-#define FLASH_M25 1 /**< Value to pass flash callbacks to use M25 Flash */
+#ifndef SWIFTNAV_FLASH_H
+#define SWIFTNAV_FLASH_H
```
fnoble: This file is now just common return codes between the two kinds of flash? (Which is fine, just checking I understand)

src/peripherals/stm_flash.c
cbeighley: @fnoble It's a callback to read something from the STM flash; seemed like the most logical place to me. Can move it to ```init.c``` though, I don't have a strong opinion about it.

@mfine It's actually not registered in the bootloader atm. We should make sure to clearly define which callbacks are registered in the bootloader vs the application somewhere.

src/flash.h
cbeighley: Yep, exactly

src/peripherals/stm_flash.c
cbeighley: Actually, I guess it's ROM, not flash. So yeah it probably makes sense to put somewhere else.

mfine: @cbeighley isn't it registered here?

https://github.com/swift-nav/piksi_bootloader/blob/master/src/flash_callbacks.c#L270-286

mfine: Ahh, maybe stm_unique_id_callback_register isn't being called anywhere in the bootloader at the moment.

fnoble: As a general principle I think its good to keep the things under `board/` and `peripheral/` factored out from the rest of the FW as much as possible.

They are kind of like mini libraries for interacting with certain subsystems. I think its much less confusing and more maintainable if they do not use SBP etc. directly. Its the job of the firmware to combine those libraries with the SBP or whatever functionality.

mfine:
re: the grouping of messages - they're more logical than physical, so if this group of operations make sense to keep together vs. what's in the bootloader, then probably shouldn't merge them. Don't think the message groupings are always clear cut, however.

src/peripherals/stm_flash.c
cbeighley: Makes sense, agreed

fnoble:
I'd say grouping the messages with the bootloader is a reasonable idea as it shows people that these messages are all supposed to be used together in the bootloader. The bootloader will be the only place they are used after this PR is in.

mfine:
But that might suggest two message groups - bootloader and firmware? :) The current bootloader message are around handshaking and jumping to the app, while the flash message are about reading, writing, programming - and getting unique id (which seems a little out of place). But yeah, could make sense to consolidate these messages down to the bootloader message.

/cc @mookerji

fnoble:
Ok, I don't feel strongly about the grouping, will let you guys duke it out :)

cbeighley:
I think ideally anything that would be a ```Firmware``` grouping would go under ```Bootloader```, and we'd make the bootloader the only way to touch the flash directly.

cbeighley:
Getting Unique ID is more of a .... ```Device``` grouping, maybe?

mfine:
Don't feel too strongly about the grouping - seems like it might make more sense. Deferring to grouping maestro @mookerji

mookerji:
Man, seems like everyone's tea has gone cold. Perhaps it's the rain, this melancholy downpour extinguishing this deep fire within us, this eternal passion for namespacing and modularization.

As for me, well, I too have weak feelings about this, although it does seem like flash and the bootloader  are logically separate things. I mean, aren't there other times we might want to write to flash that have nothing to do with bootloading the firmware and the NAP? The firmware writes out position data to the flash all the time, we wrote out the almanacs, surely there'll be a need to do such things in a request/response-style in the future ("I'll write your flash if you read mine", etc. etc.)? Perhaps in the future we'll just have application specific messages/handlers to do those things?

I'm certainly missing some context here, since I'm not working on this directly. I do think there's a fair case to be made for not making the flash accessible unless you're in the bootloader, but I think that choice needs to be made explicit in the SBP spec.

cbeighley:
I think my feeling is that external systems shouldn't be interacting with the flash directly through callbacks in the application firmware, only in the bootloader. Therefore any messages to interact with the flash directly (i.e. not through the filesystem) should be only available in the bootloader and organized in the protocol as such.

mookerji:
Sure, I totally agree about external systems and the app firmware. I will add, you could accomplish this same thing by removing the flash callbacks and simply noting in SBP's flash.yaml spec that these messages are only accessible during bootloading (it's less work, otherwise you'll have to propagate the generated code everywhere). I think this point is unrelated to this PR though (which you should go ahead and merge).

src/peripherals/stm_flash.c
cbeighley: Fixed.

src/peripherals/stm_flash.c
cbeighley: Fixed.

cbeighley:
Fixes per comments. Please review again @fnoble

src/init.c
```diff
@@ -72,6 +73,26 @@ static void reset_callback_register(void)
...
+ */
+void stm_unique_id_callback(u16 sender_id, u8 len, u8 msg[], void* context)
+{
+  (void)sender_id; (void)len; (void)msg; (void) context;
+
+  sbp_send_msg(SBP_MSG_STM_UNIQUE_ID, 12, (u8*)STM_UNIQUE_ID_ADDR);
+}
+
+/** Register callback to read Device's Unique ID. */
+void stm_unique_id_callback_register(void)
```
fnoble: Where are the prototypes for these functions?

src/init.c
```diff
@@ -72,6 +73,26 @@ static void reset_callback_register(void)
...
 }
 
+/** Callback to read STM32F4's hardcoded unique ID.
+ * Sends STM32F4 unique ID (12 bytes) back to host.
+ */
+void stm_unique_id_callback(u16 sender_id, u8 len, u8 msg[], void* context)
+{
+  (void)sender_id; (void)len; (void)msg; (void) context;
+
+  sbp_send_msg(SBP_MSG_STM_UNIQUE_ID, 12, (u8*)STM_UNIQUE_ID_ADDR);
```
fnoble: Where is `STM_UNIQUE_ID_ADDR` defined?

cbeighley: @fnoble Forgot to move it out of stm_flash.h, now it's in init.c.

src/init.c
cbeighley: @fnoble They're just used in init.c now, want me to put them in init.h?

fnoble: If they are just used in that file you can indicate it using the `static` keyword.

cbeighley: Done.

mookerji:
retest this please

fnoble:
retest this please

fnoble:
lgtm, merge at will!

PR #434, opened 2015-06-11
Use lock counter from both rover and base

fnoble:
**Why:** When several satellites drop at once causing the total number to go below 4 the filter doesn't update. When the satellites come back the filter needs to know that their integer ambiguity has changed. If the filter is unable to detect this then very large baseline outliers are generated and the filter never recovers. For base observations this was done by the lock counter mechanism but there is currently no analogue for the local observations.

**How:** Extend the lock counter mechanism to the rover as well. To do this we add a lock counter field to the single differences and make this equal to the _sum_ of the base and rover lock counters (lock counters are typically incremented to indicate a loss of lock). We now check this combined lock counter values for changes.

**Testing:** Tested on the bench by wiggling the antenna cable. Previously it was quite easy to cause a large error but with this change I have been unable to cause a large baseline outlier to occur. Further testing required on HITL as well as further manual testing.

review plz: @kovach 
fyi: @cbeighley @mookerji @denniszollo

### DIFF
--- a/src/solution.c
+++ b/src/solution.c
@@ -58,6 +58,8 @@ u32 obs_output_divisor = 2;
 double known_baseline[3] = {0, 0, 0};
 u16 msg_obs_max_size = 104;
 
+static u16 lock_counters[MAX_SATS];
+
 void solution_send_sbp(gnss_solution *soln, dops_t *dops)
 {
   if (soln) {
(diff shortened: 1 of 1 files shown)


mookerji:
mon€y

fnoble:
HITL looks good (finally)

cbeighley:
Looks good. Have you bumped libswiftnav to master?

fnoble:
Yup

cbeighley:
merge at will

PR #436, opened 2015-06-11
Update assert() to use screaming death

fnoble:
Calls to `assert()` weren't being properly output previously. Now we direct the output to `_screaming_death()`.

Tested on the bench, doesn't need HITL.

cc: @cbeighley


cbeighley:
lgtm

PR #438, opened 2015-06-12
Settings pdf

denniszollo:
update settings doc


denniszollo:
please ignore - accidental extra commits for now

denniszollo:
@cbeighley

cbeighley:
Under 3.1 ext events, it says "Requires NAP firmware ¿= 0.12."

cbeighley:
amb init var and new amb var have changed to 1e25. Can we generate these default values in the pdf programatically?

denniszollo:
definitely, but I think it is kind of hard from a dependency perspective.  Almost like the code that generates the settings would need to checkout a firmware version as part of itself.

denniszollo:
This kind of speaks to why the settings.pdf belongs in the firmware repo.

denniszollo:
Although I don't think it would be bad to procedurally enforce that when people muck with the settings they have to update the documentation .yaml for this pdf and the console.

cbeighley:
Looks good to me. Maybe the README.md should show where the settings pdf can be found?

PR #449, opened 2015-06-23
New Settings

mfine:
Callbacks for new settings messages introduced in https://github.com/swift-nav/libsbp/pull/172. Essentially, the changes here are:

1. break up the settings message into a read and write message.
2. use a done message to indicate end of settings read by index.

Most of the logic from the existing settings message was re-used, but should give this a thorough review.

/cc @fnoble @mookerji @denniszollo

### DIFF
--- a/src/settings.c
+++ b/src/settings.c
@@ -160,8 +160,9 @@ static const struct setting_type type_int = {
   int_to_string, int_from_string, NULL, NULL, &type_float,
 };
 
-static void settings_msg_callback(u16 sender_id, u8 len, u8 msg[], void* context);
 static void settings_save_callback(u16 sender_id, u8 len, u8 msg[], void* context);
+static void settings_write_callback(u16 sender_id, u8 len, u8 msg[], void* context);
+static void settings_read_callback(u16 sender_id, u8 len, u8 msg[], void* context);
 static void settings_read_by_index_callback(u16 sender_id, u8 len, u8 msg[], void* context);
 
 int settings_type_register_enum(const char * const enumnames[], struct setting_type *type)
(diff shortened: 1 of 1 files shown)


mfine:
Ready for review.

mfine:
Also potentially add checks for sender_id == 42 to ensure safety of settings.

mfine:
Added sender_id checks for read/write messages.

mfine:
@henryhallam @gsmcmullin made some changes to the settings messages in the firmware that could use a good review.

src/settings.c
```diff
@@ -275,62 +282,107 @@ static int settings_format_setting(struct setting *s, char *buf, int len)
...
-        tok++;
-        switch (tok) {
-        case 1:
-          setting = (const char *)&msg[i+1];
-          break;
-        case 2:
-          if (i + 1 < len)
-            value = (const char *)&msg[i+1];
+  if (len == 0)
+    goto error;
```
cbeighley: We've tried to avoid `goto` elsewhere, this function is the only place it's currently used - maybe just `log_error` and `return`?

src/settings.c
```diff
@@ -275,62 +282,107 @@ static int settings_format_setting(struct setting *s, char *buf, int len)
   return buflen;
 }
 
-static void settings_msg_callback(u16 sender_id, u8 len, u8 msg[], void* context)
+static void settings_write_callback(u16 sender_id, u8 len, u8 msg[], void* context)
 {
-  (void)sender_id; (void) context;
+  (void) context;
+
+  if (sender_id != 0x42)
```
cbeighley: Would it be good to `log_warn` or `log_info` when receiving a message that requires sender_id 0x42 but receives it with a different sender_id?

src/settings.c
mfine: Overall, there are 10 goto's here - so I'm thinking that this was done to reduce code verbosity. But fine with me if we want to yank it.

src/settings.c
mfine: Sounds good to me!

src/settings.c
cbeighley: Also I think single statement conditionals should still have braces: 
```
if (len == 0) {
  log_error("Error in settings write message\n");
  return;
}
```

src/settings.c
```diff
@@ -339,7 +391,10 @@ static void settings_msg_callback(u16 sender_id, u8 len, u8 msg[], void* context
 
 static void settings_read_by_index_callback(u16 sender_id, u8 len, u8 msg[], void* context)
 {
-  (void)sender_id; (void) context;
+  (void) context;
+
+  if (sender_id != 0x42)
```
denniszollo: Lets make sure to update the LIBSBP document to explain what a sender ID of 0x42 will allow on the device.  I sort of think the read functions should be allowed from any sender id, while the write should be disallowed.

mfine:
Got rid of the goto's, and added some errors around bad sender id.

cbeighley:
@mfine Maybe it would be good to have the error message be different for the wrong sender_id vs len == 0 cases so they can be distinguished.

mfine:
Changed the error message to indicate invalid sender.

mfine:
Going to bring this in.

PR #452, opened 2015-06-24
Update to new file_io messages

mfine:
Update sbp and move to new file_io messages. See https://github.com/swift-nav/libsbp/pull/183.

/cc @fnoble @denniszollo @mookerji

### DIFF
--- a/src/sbp_fileio.c
+++ b/src/sbp_fileio.c
@@ -31,13 +31,13 @@ void sbp_fileio_setup(void)
 {
   static sbp_msg_callbacks_node_t read_node;
   sbp_register_cbk(
-    SBP_MSG_FILEIO_READ,
+    SBP_MSG_FILEIO_READ_REQUEST,
     &read_cb,
     &read_node
   );
   static sbp_msg_callbacks_node_t read_dir_node;
   sbp_register_cbk(
-    SBP_MSG_FILEIO_READ_DIR,
+    SBP_MSG_FILEIO_READ_DIR_REQUEST,
     &read_dir_cb,
     &read_dir_node
   );
@@ -49,18 +49,18 @@ void sbp_fileio_setup(void)
   );
   static sbp_msg_callbacks_node_t write_node;
   sbp_register_cbk(
-    SBP_MSG_FILEIO_WRITE,
+    SBP_MSG_FILEIO_WRITE_REQUEST,
     &write_cb,
     &write_node
   );
 }
 
 /** File read callback.
- * Responds to a SBP_MSG_FILEIO_READ message.
+ * Responds to a SBP_MSG_FILEIO_READ_REQUEST message.
  *
  * Reads a certain length (up to 255 bytes) from a given offset. Returns the
- * data in a SBP_MSG_FILEIO_READ message where the message length field indicates
- * how many bytes were succesfully read.
+ * data in a SBP_MSG_FILEIO_READ_RESPONSE message where the message length field
+ * indicates how many bytes were succesfully read.
  */
 static void read_cb(u16 sender_id, u8 len, u8 msg[], void* context)
 {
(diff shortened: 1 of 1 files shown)


src/sbp_fileio.c
```diff
@@ -81,16 +81,16 @@ static void read_cb(u16 sender_id, u8 len, u8 msg[], void* context)
   len += cfs_read(f, buf + len, readlen);
   cfs_close(f);
 
-  sbp_send_msg(SBP_MSG_FILEIO_READ, len, buf);
+  sbp_send_msg(SBP_MSG_FILEIO_READ_RESPONSE, len, buf);
```
denniszollo: above here (line 77), we need a check that len + readlen < 256, otherwise the you could easily overrun buffer[256].  I don't think we can trust the sender of the message took into account that readlen needs to be correct.

mfine:
Something else to consider is whether we also want to do sender_id checks to ensure these messages are not coming from other piksi's:

```c
  if (sender_id != 0x42)
    return;
```

cbeighley:
If we are changing the IDs so there's no amibiguity as to whether a particular message is a Request or a Response, is the sender_id check necessary?

mfine:
Shouldn't be - just extra protection

extra protection

cbeighley:
Sounds good - I'm all about it.

fnoble:
Would Piksis ever want to program each other or read each other's files? Probably not but worth considering...

cbeighley:
Dennis suggested a general purpose wrapper message that could contain another message within itself. The wrapper message could also have flags for what was to be done with the wrapped message (forwarded, consumed, etc).

mookerji:
Not sure we want Piksi's playing with each other that way. We don't have quite the right protection in place,...



</gutter>

denniszollo:
The nice thing about the wrapper message is that the complexity of forwarded messages goes away in the protocol.

cbeighley:
image

```Here's my pecker.```

**Comment by mookerji** [comment]
I think the Piksi's should *cough cough* abstain from sending each other any such file messages, wrapped in anything, much less brown paper packet frames (tied up in strings).

**Comment by mfine** [comment]
I don't think we have any demonstrated need for encapsulation or piksi-to-piksi file_io, flash, or settings operations. If there's a well-defined specification and application for encapsulation, we could approach the idea then; otherwise, seems premature. For going out in 1.0, seems like a good idea to lock these things down a little and remove inadvertent updates.

File: src/sbp_fileio.c (hunk 1 above)
Comment (mfine) [inline src/sbp_fileio.c]: `len` is an 8-bit value, so you can't overrun the buffer - it'll just wrap. If you try and in a check for this, the compiler is warning:

```
sbp_fileio.c:86:3: error: comparison is always false due to limited range of data type
```

**Comment by mfine** [comment]
Added the `sender_id` checks.

File: src/sbp_fileio.c [hunk 2]
```diff
@@ -81,25 +83,27 @@ static void read_cb(u16 sender_id, u8 len, u8 msg[], void* context)
   len += cfs_read(f, buf + len, readlen);
```
denniszollo: Let's say the original message is 16 bits long as sent (len=16).  Lets say the readlen field of the message is it's max value (256).  the cfs_read operation just overran buf.

mfine: Ah, the `u8` len + the `u8` readlen in the call to `cfs_read` could overwrite buf 256 bytes. Added MIN(readlen, 255-len) to determine how much to read into the buffer. Thanks!

src/sbp_fileio.c
```diff
@@ -49,31 +50,33 @@ void sbp_fileio_setup(void)
...
- * how many bytes were succesfully read.
+ * data in a SBP_MSG_FILEIO_READ_RESPONSE message where the message length field
+ * indicates how many bytes were succesfully read.
  */
 static void read_cb(u16 sender_id, u8 len, u8 msg[], void* context)
 {
-  (void)sender_id;
   (void)context;
 
+  if (sender_id != 0x42)
```
mookerji: I may have missed this earlier, but sender ID 42 is used by those host computer writing to the flash? This would useful to have defined in a constant (and comment) somewhere in the header.

mfine:
@henryhallam @gsmcmullin minor changes here around the file io messaging paths - could use a brief review.

mfine:
Rebased.

mfine:
Bringing this in.

Issue #246, opened 2015-02-14
Add more NMEA strings


fnoble:
See:

https://groups.google.com/d/msg/swiftnav-discuss/ozrLQJOjpWo/-Xq1mLd1DewJ

We should at least add GPRMC and GPVTG.

PR #458, opened 2015-06-26
NMEA messages

### LINKED ISSUE #443: GPGSV needs updating after CN0 change
Units need re-scaling.

soulcmdc:
Fixes #443, #246.
Still needs testing, but would like to get some eyes on the code early.

### DIFF
--- a/src/nmea.c
+++ b/src/nmea.c
@@ -16,6 +16,7 @@
 #include <string.h>
 
 #include <libswiftnav/coord_system.h>
+#include <libswiftnav/constants.h>
 
 #include "board/nap/track_channel.h"
 #include "nmea.h"
@@ -25,6 +26,10 @@
 
 const char NMEA_MODULE[] = "nmea";
 
+u32 gpgsv_msg_rate = 10;
+u32 gprmc_msg_rate = 10;
+u32 gpvtg_msg_rate = 10;
+u32 gpgll_msg_rate = 10;
 static struct nmea_dispatcher *nmea_dispatchers_head;
 /** \addtogroup io
  * \{ */
@@ -62,6 +67,15 @@ void nmea_output(char *s)
 
   for (struct nmea_dispatcher *d = nmea_dispatchers_head; d; d = d->next)
     d->send(s);
+}
+
+void nmea_setup(void)
+{
+
+  SETTING("nmea", "gpgsv_msg_rate", gpgsv_msg_rate, TYPE_INT);
+  SETTING("nmea", "gprmc_msg_rate", gprmc_msg_rate, TYPE_INT);
+  SETTING("nmea", "gpvtg_msg_rate", gpvtg_msg_rate, TYPE_INT);
+  SETTING("nmea", "gpgll_msg_rate", gpgll_msg_rate, TYPE_INT);
 }
 
 /** Calculate the checksum of an NMEA sentence.
@@ -104,10 +118,10 @@ void nmea_gpgga(const double pos_llh[3], const gps_time_t *gps_t, u8 n_used,
 
   double frac_s = fmod(gps_t->tow, 1.0);
 
-  s16 lat_deg = (s16)((180.0 / M_PI) * pos_llh[0]);
-  double lat_min = fabs(60 * ((180.0 / M_PI) * pos_llh[0] - lat_deg));
-  s16 lon_deg = (s16)((180.0 / M_PI) * pos_llh[1]);
-  double lon_min = fabs(60 * ((180.0 / M_PI) * pos_llh[1] - lon_deg));
+  s16 lat_deg = R2D * (pos_llh[0]);
+  double lat_min = MINUTES(pos_llh[0]);
+  s16 lon_deg = R2D * (pos_llh[1]);
+  double lon_min = MINUTES(pos_llh[1]);
   lat_deg = abs(lat_deg);
   lon_deg = abs(lon_deg);
 
@@ -193,9 +207,9 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
         bufp += sprintf(
           bufp, ",%02d,%02d,%03d,%02d",
           nav_meas[n].prn + 1,
-          (u8)round(el * 180.0 / M_PI),
-          (u16)round(az * 180.0 / M_PI),
-          (u8)(10.0 * nav_meas[n].snr)
+          (u8)round(el * R2D),
+          (u16)round(az * R2D),
+          (u8)round(nav_meas[n].snr)
           );
       } else {
         bufp += sprintf(bufp, ",,,,");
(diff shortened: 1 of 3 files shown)


soulcmdc:
cc @fnoble ? Not sure who else to include

src/nmea.c
```diff
@@ -209,6 +209,142 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
...
 }
 
+/** Assemble an NMEA GPRMC message and send it out NMEA USARTs.
+ * NMEA RMC contains minimum GPS data 
+ *
+ * \param nav_meas Pointer to navigation_measurement struct.
+ * \param soln Pointer to gnss_solution struct
+ */
+void nmea_gprmc(const navigation_measurement_t *nav_meas,
+                  const gnss_solution *soln, const gps_time_t *gps_t)
```
fnoble: nit: Indentation

src/nmea.c
```diff
@@ -209,6 +209,142 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
...
+
+  /* Conversion to magnitue knotts */
+  velocity = sqrt(x*x + y*y + z*z)*1.94385;
+
+  double az, el;
+  wgsecef2azel(nav_meas[0].sat_pos, soln->pos_ecef, &az, &el);
+
+  char buf[80];
+  u8 n = sprintf(buf,
+                "$GPRMC,%02d%02d%06.3f,A," // Command, Time (UTC), Valid
```
fnoble: `/* */` not `//`

fnoble:
Looking great, we might want to consider if we want to output all the messages all the time or not. One thing we could have is a setting for each message that sets how frequently each message will be sent. The setting could be a decimation factor, i.e. the `solution_send_nmea()` function gets called at the standard solution rate and then you can specify if you want each message to be output never, every time, every second time etc. Just an idea.

@denniszollo : I know you were interested in this setting too right?

denniszollo:
Yes I think we need rates on this stuff before we have the message white list

denniszollo:
maybe if the rate divisor is 0 we don't send at all

soulcmdc:
@fnoble @denniszollo 
I think this is good to go.
Tested with:
1. OpenCPN  (http://opencpn.org/ocpn/) 
2. http://freenmea.net/decoder (By far the stricter parser. Found several small issues here.)

I'm using rate_periods instead of divisors.
GPGGA is still sent with solution_send_baseline() more frequently.

soulcmdc:
SNR is now matching console, and similar to other units out there.

soulcmdc:
We only send out those messages when we have observations that are valid.  I think even in simulation mode we want to consider the messages valid so we can use those messages, so I didn't do anything more complex than just populate the field.

src/nmea.c
```diff
@@ -209,6 +209,163 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
...
+
+  /* Conversion to magnitue knotts */
+  vknotts = sqrt(x*x + y*y + z*z)*1.94385;
+  /* Conversion to magnitue km/hr */
+  vkmhr = sqrt(x*x + y*y + z*z) * (3600/1000); 
+
+  char buf[80];
+  u8 n = sprintf(buf, 
+                  "$GPVTG,%05.1f,T," /* Command, course, */
+                  ",M," /* Magnetic Course (omitted) */
```
denniszollo: It's nice that they let you omit things.  It would suck to have to maintain and update a table of declination.

src/nmea.c
```diff
@@ -209,6 +209,163 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
...
+
+  double az, el;
+  wgsecef2azel(nav_meas[0].sat_pos, soln->pos_ecef, &az, &el);
+
+  float vknotts, vkmhr;
+  float x,y,z;
+  x = soln->vel_ned[0];
+  y = soln->vel_ned[1];
+  z = soln->vel_ned[2];
+  float course = atan2(y,x);
```
denniszollo: Is there any chance that x and/or y are exactly zero?  Will this fail if atan2 returns NAN?

src/nmea.c
```diff
@@ -209,6 +209,163 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
...
+
+  /* Conversion to magnitue knotts */
+  vknotts = sqrt(x*x + y*y + z*z)*1.94385;
```
denniszollo: Can we use the preprocessor perhaps to define these constants?

src/nmea.c
```diff
@@ -209,6 +209,163 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
...
+   */ 
+  time_t unix_t;
+  struct tm t;
+
+  unix_t = gps2time(*gps_t);
+  gmtime_r(&unix_t, &t);
+
+  double frac_s = fmod(gps_t->tow, 1.0);
+  s16 lat_deg = (s16)((180.0 / M_PI) * soln->pos_llh[0]);
+  double lat_min = fabs(60 * ((180.0 / M_PI) * soln->pos_llh[0] - lat_deg));
```
denniszollo: Might look cleaner of we used RAD2DEG instead of all these divisions and parens.

src/nmea.h
```diff
@@ -37,6 +37,11 @@ void nmea_gpgga(const double pos_llh[3], const gps_time_t *gps_t, u8 n_used,
...
 void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
                 const gnss_solution *soln);
+void nmea_gprmc(const navigation_measurement_t *nav_meas,
+                  const gnss_solution *soln, const gps_time_t *gps_t);
+void nmea_gpvtg(const navigation_measurement_t *nav_meas,
+                const gnss_solution *soln);
+void nmea_gpgll(const gnss_solution *soln, const gps_time_t *gps_t);
 
 /** Register a new dispatcher for NMEA messages
  *
```
denniszollo: If we use preprocessor the constants and conversion factors might go here.

src/solution.c
```diff
@@ -54,6 +54,10 @@ systime_t last_dgnss;
 
 double soln_freq = 10.0;
 u32 obs_output_divisor = 2;
+u32 gpgsv_msg_period = 10;
```
denniszollo: do we think "period: is clear as for the name of these parameters?  I don't want anyone to think that these are the "seconds" between messages.  It might make more sense to have them be called rates or output divisors.  Other mfctrs seem to call them rates or rate divisors.

src/solution.c
```diff
@@ -746,6 +771,11 @@ void solution_setup()
 
   SETTING("sbp", "obs_msg_max_size", msg_obs_max_size, TYPE_INT);
```
denniszollo: I'd like to make sure that these names are clear as above.  I'd rather see gpgsv_output_divisor or something as the naming pattern.

denniszollo:
So I think this looks good.  Stylistically I'd like to see use of preprocessor for some of the constants and I'd like to consider alternate names for the output divisor variables on each NMEA message.

src/nmea.c
denniszollo: I worry that our solution algorithm might return zero for these on its first output.

src/solution.c
soulcmdc: Rate is probably more appropriate and with fewer connotations.

src/nmea.c
soulcmdc: So far i haven't seen that.  We only output these when there's a solution.

src/nmea.c
soulcmdc: Fair point, added RAD2DEG_INT, RAD2DEG_FLOAT, MINUTES, MS2KNOTTS, MS2KMHR preprocessor commands.

soulcmdc:
Addressed/rebutted all comments.  Added the preprocessor directives (it does make things cleaner), period->rate.

src/nmea.c
```diff
@@ -195,7 +195,7 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
           nav_meas[n].prn + 1,
           (u8)round(el * 180.0 / M_PI),
           (u16)round(az * 180.0 / M_PI),
-          (u8)(10.0 * nav_meas[n].snr)
+          (u8)(nav_meas[n].snr)
```
fnoble: `(u8)round(nav_meas[n].snr)`?

src/nmea.c
fnoble: `atan2(0, 0) == 0` by convention, so this is fine

src/nmea.c
```diff
@@ -209,6 +209,163 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
...
+  char lon_dir = soln->pos_llh[1] < 0 ? 'W' : 'E';
+
+  float velocity;
+  float x,y,z;
+  x = soln->vel_ned[0];
+  y = soln->vel_ned[1];
+  z = soln->vel_ned[2];
+  float course = atan2(y,x);
+
+  /* Conversion to magnitue knotts */
```
fnoble: s/knott/knot/ everywhere

src/nmea.h
```diff
@@ -32,11 +32,23 @@
 #define NMEA_GGA_FIX_MANUAL  7
 #define NMEA_GGA_FIX_SIM     8
 
+#define RAD2DEG_INT(X) (s16)((180.0 / M_PI) * X)
+#define RAD2DEG_FLOAT(X) ((180.0/M_PI) * X)
```
fnoble: we also have these already defined:
https://github.com/swift-nav/libswiftnav/blob/master/include/libswiftnav/constants.h#L25

src/nmea.h
```diff
@@ -32,11 +32,23 @@
...
+#define MINUTES(X) fabs(60 * (RAD2DEG_FLOAT(X) - RAD2DEG_INT(X)));
+#define MS2KNOTTS(x,y,z) sqrt(x*x + y*y + z*z) * 1.94385
+#define MS2KMHR(x,y,z) sqrt(x*x + y*y + z*z) * (3600/1000)
```
fnoble: These would eventually be candidates to move to libswiftnav. Also you should put parens around macro function arguments to prevent strangeness with operator precedence:

`#define MS2KNOTTS(x,y,z) sqrt((x)*(x) + (y)*(y) + (z)*(z)) * 1.94385`

src/nmea.h
```diff
@@ -32,11 +32,23 @@
...
+
+
 void nmea_gpgga(const double pos_llh[3], const gps_time_t *gps_t, u8 n_used,
                 u8 fix_type, double hdop);
 void nmea_gpgsa(const tracking_channel_t *chans, const dops_t *dops);
 void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
                 const gnss_solution *soln);
+void nmea_gprmc(const navigation_measurement_t *nav_meas,
+                  const gnss_solution *soln, const gps_time_t *gps_t);
```
fnoble: indentation

src/solution.c
```diff
@@ -54,6 +54,10 @@ systime_t last_dgnss;
 
 double soln_freq = 10.0;
 u32 obs_output_divisor = 2;
+u32 gpgsv_msg_rate = 10;
+u32 gprmc_msg_rate = 10;
+u32 gpvtg_msg_rate = 10;
+u32 gpgll_msg_rate = 10;
```
fnoble: I'd be tempted to move these to `nmea.c`, `solution.c` is getting bloated as is.

src/solution.c
```diff
@@ -105,9 +109,30 @@ void solution_send_nmea(gnss_solution *soln, dops_t *dops,
                fix_mode, dops->hdop);
   }
 
-  DO_EVERY(10,
```
fnoble: Could move the meat of this function into a function in `nmea.c`

src/solution.c
```diff
@@ -746,6 +771,11 @@ void solution_setup()
 
   SETTING("sbp", "obs_msg_max_size", msg_obs_max_size, TYPE_INT);
 
+  SETTING("nmea", "gpgsv_msg_rate", gpgsv_msg_rate, TYPE_INT);
+  SETTING("nmea", "gprmc_msg_rate", gprmc_msg_rate, TYPE_INT);
+  SETTING("nmea", "gpvtg_msg_rate", gpvtg_msg_rate, TYPE_INT);
+  SETTING("nmea", "gpgll_msg_rate", gpgll_msg_rate, TYPE_INT);
```
fnoble: Again, could move these to a setup function in nmea.c (if it doesn't have one already)

src/nmea.c
soulcmdc: Yep

src/nmea.c
soulcmdc: d'oh

src/nmea.c
```diff
@@ -104,10 +118,10 @@ void nmea_gpgga(const double pos_llh[3], const gps_time_t *gps_t, u8 n_used,
 
   double frac_s = fmod(gps_t->tow, 1.0);
 
-  s16 lat_deg = (s16)((180.0 / M_PI) * pos_llh[0]);
-  double lat_min = fabs(60 * ((180.0 / M_PI) * pos_llh[0] - lat_deg));
-  s16 lon_deg = (s16)((180.0 / M_PI) * pos_llh[1]);
-  double lon_min = fabs(60 * ((180.0 / M_PI) * pos_llh[1] - lon_deg));
+  s16 lat_deg = (s16)R2D * (pos_llh[0]);
```
fnoble: Did you test this again? I think the casting will cause problems here. I think it should be:
`s16 lat_deg = R2D * pos_llh[0];`

src/nmea.c
```diff
@@ -104,10 +118,10 @@ void nmea_gpgga(const double pos_llh[3], const gps_time_t *gps_t, u8 n_used,
...
+  double lat_min = MINUTES(pos_llh[0]);
+  s16 lon_deg = (s16)R2D * (pos_llh[1]);
```
fnoble: Check cast

src/nmea.c
```diff
@@ -209,6 +223,192 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
...
+   * currently
+   */
+  time_t unix_t;
+  struct tm t;
+
+  unix_t = gps2time(*gps_t);
+  gmtime_r(&unix_t, &t);
+  double frac_s = fmod(gps_t->tow, 1.0);
+
+  s16 lat_deg = (s16)R2D * (soln->pos_llh[0]);
```
fnoble: Check cast

src/nmea.c
```diff
@@ -209,6 +223,192 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
...
+  double lat_min = MINUTES(soln->pos_llh[0]);
+  s16 lon_deg = (s16)R2D * (soln->pos_llh[1]);
```
fnoble: Check cast

src/nmea.c
```diff
@@ -209,6 +223,192 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
...
+  char buf[100];
+  u8 n = sprintf(buf,
+                "$GPRMC,%02d%02d%06.3f,A," /* Command, Time (UTC), Valid */
+                "%02d%010.7f,%c,%03d%010.7f,%c," /* Lat/Lon */
+                "%06.2f,%05.1f," /* Speed, Course */
+                "%02d%02d%02d," /* Date Stamp */
+                ",", /* Variation */
+                t.tm_hour, t.tm_min, t.tm_sec + frac_s,
+                lat_deg, lat_min, lat_dir, lon_deg, lon_min, lon_dir,
+                velocity, course * 180.0 / M_PI,
```
fnoble: use R2D here too?

src/nmea.c
```diff
@@ -209,6 +223,192 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
...
+  vknots = MS2KNOTTS(x,y,z);
+  /* Conversion to magnitue km/hr */
+  vkmhr = MS2KMHR(x,y,z);
+
+  char buf[80];
+  u8 n = sprintf(buf, 
+                  "$GPVTG,%05.1f,T," /* Command, course, */
+                  ",M," /* Magnetic Course (omitted) */
+                  "%06.2f,N,%06.2f,K", /* Speed (knots, km/hr) */
+                  course*180.0 / M_PI,
```
fnoble: R2D

src/nmea.c
```diff
@@ -209,6 +223,192 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
...
+   *          LAT                  UTC
+   */ 
+  time_t unix_t;
+  struct tm t;
+
+  unix_t = gps2time(*gps_t);
+  gmtime_r(&unix_t, &t);
+
+  double frac_s = fmod(gps_t->tow, 1.0);
+  s16 lat_deg = (s16)R2D * (soln->pos_llh[0]);
```
fnoble: Check cast

src/nmea.c
```diff
@@ -209,6 +223,192 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
...
+  double lat_min = MINUTES(soln->pos_llh[0]);
+  s16 lon_deg = (s16)R2D * (soln->pos_llh[1]);
```
fnoble: Check cast

src/nmea.h
```diff
@@ -32,11 +32,24 @@
 #define NMEA_GGA_FIX_MANUAL  7
 #define NMEA_GGA_FIX_SIM     8
 
+#define MINUTES(X) fabs(60 * (R2D * (X) - ((s16)R2D * (X))));
```
fnoble: Check cast here too, should be:
`#define MINUTES(X) fabs(60 * (R2D * (X) - ((s16)(R2D * (X)))));`

src/nmea.c
```diff
@@ -209,6 +223,192 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
...
+    nmea_gprmc(nm, soln, &soln->time);
+  );
+  DO_EVERY(gpvtg_msg_rate,
+    nmea_gpvtg(nm, soln);
+  );
+  DO_EVERY(gpgll_msg_rate,
+    nmea_gpgll(soln, &soln->time);
+  );
+
+}
```
fnoble: newline

soulcmdc:
@fnoble Fixed comments, and tested.  Things look good.

soulcmdc:
@fnoble All cleaned up.

Issue #238, opened 2015-01-28
Solution thread calls calc_PVT with zero sats once on startup


fnoble:
This leads to the PVT failed to converge message

PR #466, opened 2015-06-30
calc_pvt called with 0 sats on startup

ungureanuvladvictor:
Fixes https://github.com/swift-nav/piksi_firmware/issues/238

/cc @fnoble

### DIFF
--- a/src/solution.c
+++ b/src/solution.c
@@ -314,7 +314,7 @@ static void timer_set_period_check(uint32_t timer_peripheral, uint32_t period)
   __asm__("CPSIE i;");
 }
 
-static void simulation_thread()
+static void solution_simulation()
 {
   /* Set the timer period appropriately. */
   timer_set_period_check(TIM5, round(65472000 * (1.0/soln_freq)));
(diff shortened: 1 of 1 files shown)


src/solution.c
```diff
@@ -359,6 +410,10 @@ static msg_t solution_thread(void *arg)
       memcpy(nav_meas_old, nav_meas, sizeof(nav_meas));
       n_ready_old = n_ready;
 
+      if (n_ready_tdcp == 0) {
+        return 0;
```
fnoble: hmm, should this actually be `continue`?

fnoble: can also change the `     if (n_ready >= 4) {` check to do the same thing.

src/solution.c
```diff
@@ -314,6 +314,50 @@ static void timer_set_period_check(uint32_t timer_peripheral, uint32_t period)
   __asm__("CPSIE i;");
 }
 
+static void simulation_thread()
```
fnoble: maybe pick a different name, e.g. `solution_simulation` or something. `_thread` to me implies this is its own thread but really its running from the solution thread.

fnoble:
How about making this line an early continue too:

https://github.com/swift-nav/piksi_firmware/pull/466/files#diff-a8b8debb0aca96bb1c1d74a97c2afaffR391

Also, what testing have you done on this?

ungureanuvladvictor:
Just bench testing. Simulation mode works. Needs HITL now.

fnoble:
HITL looks good!

PR #467, opened 2015-07-01
Sanity check for surveyed position.

ungureanuvladvictor:
I'm not really sure what the distance threshold should be. 

/cc: @denniszollo @fnoble 

### DIFF
--- a/src/system_monitor.c
+++ b/src/system_monitor.c
@@ -18,6 +18,8 @@
 #include <libsbp/version.h>
 #include <libswiftnav/logging.h>
 #include <libswiftnav/dgnss_management.h>
+#include <libswiftnav/linear_algebra.h>
+#include <libswiftnav/coord_system.h>
 
 #include "board/nap/nap_common.h"
 #include "board/max2769.h"
@@ -28,9 +30,15 @@
 #include "manage.h"
 #include "simulator.h"
 #include "system_monitor.h"
+#include "position.h"
 
 #define WATCHDOG_HARDWARE_PERIOD_MS 30000  /* Actual period may vary +88% -32% */
 #define WATCHDOG_THREAD_PERIOD_MS 15000
+
+/* Maximum distance between calculated and surveyed base station single point 
+ * position for error checking.
+ */
+#define BASE_STATION_DISTANCE_THRESHOLD 15000
 
 /* Time between sending system monitor and heartbeat messages in milliseconds */
 static uint32_t heartbeat_period_milliseconds = 1000;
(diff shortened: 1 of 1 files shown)


src/system_monitor.c
```diff
@@ -28,10 +30,13 @@
 #include "manage.h"
 #include "simulator.h"
 #include "system_monitor.h"
+#include "position.h"
 
 #define WATCHDOG_HARDWARE_PERIOD_MS 30000  /* Actual period may vary +88% -32% */
 #define WATCHDOG_THREAD_PERIOD_MS 15000
 
+#define BASE_STATION_DISTANCE_THRESHOLD 15000
```
fnoble: Comment with units?

src/system_monitor.c
```diff
@@ -161,8 +166,23 @@ static msg_t system_monitor_thread(void *arg)
...
+
+      llhdeg2rad(base_llh, tmp);
+      wgsllh2ecef(tmp, base_ecef);
+
+      vector_subtract(3, base_ecef, position_solution.pos_ecef, tmp);
+      base_distance = vector_norm(3, tmp);
+
+      if (base_distance > BASE_STATION_DISTANCE_THRESHOLD) {
+        log_warn("Invalid surveyed position coordinates!\n");
+      }
```
fnoble: Linux Kernely style says use `} else {`

src/system_monitor.c
```diff
@@ -161,8 +167,22 @@ static msg_t system_monitor_thread(void *arg)
...
+      double base_distance;
+
+      llhdeg2rad(base_llh, tmp);
+      wgsllh2ecef(tmp, base_ecef);
+
+      vector_subtract(3, base_ecef, position_solution.pos_ecef, tmp);
+      base_distance = vector_norm(3, tmp);
+
+      if (base_distance > BASE_STATION_DISTANCE_THRESHOLD) {
+        log_warn("Invalid surveyed position coordinates!\n");
```
denniszollo: Nit- no need for ! character.

src/system_monitor.c
```diff
@@ -28,10 +30,14 @@
 #include "manage.h"
 #include "simulator.h"
 #include "system_monitor.h"
+#include "position.h"
 
 #define WATCHDOG_HARDWARE_PERIOD_MS 30000  /* Actual period may vary +88% -32% */
 #define WATCHDOG_THREAD_PERIOD_MS 15000
 
+/* Distance in meters between rover and surveyed base station position */
```
cbeighley: Isn't this the distance between the surveyed and calculated base station position? In which case something like
```
/* Maximum distance between calculated and surveyed base station single point position for error checking. */
```

cbeighley:
lgtm

src/system_monitor.c
denniszollo: I assume these are meters?
so the position must be within 15 Kilomters?

ungureanuvladvictor: Yes

PR #470, opened 2015-07-02
Verify that loaded almanac is valid (within a certain week number range)

denniszollo:
Invalidate almanac in memory if invalid for a particular PRN when discovered.

This is a check that is most relevant in the future when we are decoding the almanac or sending it to the device.  It may affect people who haven't used Piksi for a long time but used it last when we used to send the almanac to the device through the console.

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -12,6 +12,7 @@
 
 #include <math.h>
 #include <string.h>
+#include <stdlib.h>
 
 #include <ch.h>
 
@@ -96,8 +97,9 @@ static void almanac_callback(u16 sender_id, u8 len, u8 msg[], void* context)
   (void)sender_id; (void)len; (void) context;
 
   almanac_t *new_almanac = (almanac_t*)msg;
-
-  log_info("Received alamanc for PRN %02d", new_almanac->prn);
+  /* TODO: use a gpstime struct inside almanac structure */
+  log_info("Received alamanc for PRN %02d with week %04d", new_almanac->prn,
++           new_almanac->week);
   memcpy(&almanac[new_almanac->prn-1], new_almanac, sizeof(almanac_t));
 
   int fd = cfs_open("almanac", CFS_WRITE);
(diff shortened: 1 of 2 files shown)


src/manage.c
```diff
@@ -171,8 +183,16 @@ static void manage_calc_almanac_scores(void)
       /* No almanac or position/time information, give it the benefit of the
        * doubt. */
       acq_prn_param[prn].score[ACQ_HINT_ALMANAC] = 0;
+    } else if (almanac[prn].valid && time_quality > TIME_COARSE &&
+              (abs(almanac[prn].week - t.wn%1024) > MAX_ALM_WEEK_DIFF)) {
```
henryhallam: Why not roll this into the previous condition, and simply not use the almanac for this PRN rather than labeling the almanac entry as invalid?

henryhallam: Also we ought to make sure the week number stuff is thoroughly thought through.  IIRC the policy was to use full-length, non-wrapping week numbers internally, and correct the received mod-1024 value to the full length value at the time we receive a TOW, ephemeris or almanac from the sky or the internet.  So the % 1024 here, and the -1024 later, might not be appropriate here.  @fnoble is that what you had in mind?

denniszollo: I was thinking that we only need to check the file/memory once.
If we see that it is invalid, it is invalid for eterinity.

fnoble: The policy is definitely to correct the week number for rollover at the point of decoding, and use the full week number everywhere internally

denniszollo: so does this mean that our almanac file and corresponding SBP message should be the full week number?

henryhallam: yeah

fnoble: Spaces around ` % `

fnoble: Comment "TODO: Fix almanac week numbers and use `gpsdifftime()`"

denniszollo:
Rebase, add TODOs, create the following issues for next steps:

sashamitelman:
Do we check NMCT? If not, the method described in this article may be worth implementing:

PR #471, opened 2015-07-02
tracking: Send IQ correlations if enabled.

### LINKED ISSUE #460: Add implementation of tracking I/Q output setting.
The setting slipped in on the back of another pull request, but the implementation never made it in.  This requires a message to be added to libsbp.

gsmcmullin:
To fix #460.

### DIFF
--- a/src/track.c
+++ b/src/track.c
@@ -317,6 +317,17 @@ void tracking_channel_update(u8 channel)
         cs2[i].I = cs[2-i].I;
         cs2[i].Q = cs[2-i].Q;
       }
+
+      /* Output I/Q correlations using SBP if enabled for this channel */
+      if (chan->output_iq && (chan->int_ms > 1)) {
+        msg_tracking_iq_t msg = {
+          .channel = channel,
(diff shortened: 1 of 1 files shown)


src/track.c
```diff
@@ -289,6 +289,17 @@ void tracking_channel_update(u8 channel)
...
         cs2[i].Q = cs[2-i].Q;
       }
+
+      /* Output I/Q correlations using SBP if enabled for this channel */
+      if (chan->output_iq && (chan->int_ms > 1)) {
+        msg_tracking_iq_t msg = {
+          .channel = channel,
+          .sid = chan->prn,
+        };
+        memcpy(msg.corrs, chan->cs, sizeof(chan->cs));
```
fnoble: This could be a little brittle if the types of `cs` or `corrs` changes. What about assigning element by element? Another option would be to check with an `assert(sizeof(msg.corrs) == sizeof(chan->cs))`.

fnoble:
lgtm

PR #472, opened 2015-07-02
acq: Improve logging of timeouts.

gsmcmullin:
Sample load timeout lowered to info. This doesn't indicate a problem.
Other timeouts changed to error.  They probably are a result of a bug.


fnoble:
lgtm

PR #473, opened 2015-07-02
acq: Catch NaN SNR

gsmcmullin:
Complain loudly, and set to zero before it upsets other systems.

/cc @cbeighley @denniszollo


henryhallam:
Looks legit but should it be 'warning' rather than 'error'?  Do we even have a policy for the meaning of the severity levels? I've been going by:
Info = not necessarily a problem at all
Warning = probably a problem, maybe a bug, recoverable or unlikely to affect results
Error = Thing that really should not happen, leading to untrustworth Solutions / obs outputs

gsmcmullin:
This should probably also check of Infinity or zero in `power_acc`.

fnoble:
lgtm, @gsmcmullin do you want to do the check on `power_acc`? I don't think its critical...

cbeighley:
Closing as these changes are in https://github.com/swift-nav/piksi_firmware/pull/487

PR #476, opened 2015-07-06
use_tracking_channel(): use libswiftnav ephemeris_good() to reject most bad TOWs

henryhallam:
Bad tracking channel TOWs occasionally creep in.  I'm working on a separate overhaul of the nav bit parsing to take care of that, but it's also sensible to check that the TOW is within a few hours of the ephemeris TOE/TOC.  We were checking that already, but merely printing a warning if it were the case, and using the resulting bogus satellite state anyway.  With this check, we'll exclude a channel from the set of obs used for SPP and RTK if its TOW is too far out of whack wrt the ephemeris.

cc @fnoble @mookerji @denniszollo

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -483,11 +483,7 @@ static void manage_track()
 
 s8 use_tracking_channel(u8 i)
 {
-  return (tracking_channel[i].state == TRACKING_RUNNING)
-      /* Check ephemeris is useable. */
-      /* TODO: Use libswiftnav function here. */
-      && (es[tracking_channel[i].prn].valid == 1)
-      && (es[tracking_channel[i].prn].healthy == 1)
+  if ((tracking_channel[i].state == TRACKING_RUNNING)
       /* Check SNR has been above threshold for the minimum time. */
       && (tracking_channel[i].update_count
             - tracking_channel[i].snr_below_threshold_count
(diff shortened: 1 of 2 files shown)


src/manage.c
```diff
@@ -498,14 +494,24 @@ s8 use_tracking_channel(u8 i)
...
+      && (tracking_channel_snr(i) >= track_cn0_threshold)) {
+
+    /* Check ephemeris is usable. */
+    gps_time_t t = {
+      /* TODO: the following makes the week number check tautological -
+         see issue #475 */
+      .wn = es[tracking_channel[i].prn].toe.wn,
+      .tow = 1e-3 * tracking_channel[i].TOW_ms
+    };
+    return ephemeris_good(&es[tracking_channel[i].prn], t);
```
mookerji: I'm not super-familiar with the code flow following `use_tracking_channel`, but isn't there a place where this sort of already happens, `ephemeris_new(...)` in src/ephemeris.c? It seems like this logic could be consolidated there.

henryhallam: Those are two different things, really.  ephemeris_new() is checking that a newly-received ephem is consistent with a previously-known ephemeris (and might conceivably do other kinds of validation on it), before loading it into the table of active ephemerides.  This set of conditionals is more about checking that the tracking channel is in a state to produce usable observations.  The call to ephemeris_good() does two things: check that there is a valid ephemeris in the table for that PRN, and (the purpose of this PR) checking that the TOW being output from the tracking channel is not too far away from the epoch of the ephemeris.

mookerji: Ah, ok. Does the `use_tracking_channel == False` have the practical consequence of attempting to decode that ephemeris again?

henryhallam: No

fnoble:
lgtm

PR #482, opened 2015-07-06
Complain repeatedly if the NAP firmware version is incompatible

henryhallam:
cc @denniszollo 
Unfortunately getting the heartbeat message to send at this point is a bit tricky.  Is that necessary so that host software knows the SBP protocol version?  We could probably send a precanned heartbeat msg (separate from the usual heartbeat thread) if that would work.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -160,11 +160,11 @@ int main(void)
   /* Check we are running a compatible version of the NAP firmware. */
   const char *required_nap_version = "v0.12";
   if (compare_version(nap_version_string, required_nap_version) < 0) {
-    log_error("NAP firmware version newer than %s required, please update!\n"
+    while (1) {
+      log_error("NAP firmware version newer than %s required, please update!\n"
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -160,11 +160,11 @@ int main(void)
   /* Check we are running a compatible version of the NAP firmware. */
   const char *required_nap_version = "v0.12";
   if (compare_version(nap_version_string, required_nap_version) < 0) {
-    log_error("NAP firmware version newer than %s required, please update!\n"
+    while (1) {
+      log_error("NAP firmware version newer than %s required, please update!\n"
               "(instructions can be found at http://docs.swift-nav.com/)\n",
```
fnoble: Indentation

fnoble:
lgtm

fnoble:
we should probably bump the required version as well, as there have been important fixes

henryhallam:
indentation fixed; v0.14 required; rebased. merge when ready.

PR #483, opened 2015-07-07
random: better checks and thread safety on RNG.

gsmcmullin:
Check and clear error flags from RNG, restarting RNG in the case of seed error.
Report seed error and too many iterations.
Protect RNG access with a mutex.

### DIFF
--- a/src/peripherals/random.c
+++ b/src/peripherals/random.c
@@ -13,8 +13,16 @@
 #include <libopencm3/stm32/rcc.h>
 #include <libopencm3/stm32/f4/rng.h>
 
+#include <ch.h>
+
 #include "libswiftnav/common.h"
+#include "libswiftnav/logging.h"
 #include "random.h"
+
+
+#define RANDOM_MAX_TRIES 100
+
+static MUTEX_DECL(rng_mutex);
 
 /* These two methods copied from:
  * https://raw.githubusercontent.com/libopencm3/libopencm3-examples/058298fd78c23639a1a135871e6c363c9d6153f6/examples/stm32/f4/stm32f4-discovery/random/random.c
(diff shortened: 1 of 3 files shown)


src/peripherals/random.c
```diff
@@ -34,17 +42,37 @@ void rng_setup(void)
...
+      }
+      /* Write back as zero to clear */
+      RNG_SR = ~(sr & (RNG_SR_SEIS | RNG_SR_CEIS));
     }
+    /* Check if data is ready. */
+    if ((RNG_SR & RNG_SR_DRDY) == 0)
+      continue;
+
+    new_value = RNG_DR;
+  } while ((new_value == last_value) && (tries++ < RANDOM_MAX_TRIES));
```
fnoble: Is there no way to tell when the RNG has updated?

gsmcmullin: It's supposed to be `RNG_SR_DRDY`, but the datasheet says explicitly that you should check each value is different from the previous value read.

gsmcmullin: This logic here should be the same as before.  The problem with the old code is that the error conditions were never reset.

fnoble: It doesn't seem to me like `RNG_SR_DRDY` is really being used here, and the timeout seems a bit fragile as its related to CPU cycles rather than anything with the peripheral?

gsmcmullin: Maybe the timeout isn't needed.  On the failure I've seen, `RNG_SR_SEIS` was set, and recovery from that is documented.  I don't see any other reason we should get stuck in here, but if we do it means a watchdog reset.  The way it is now is a mistake, it should only increment `tries` when it actually reads `RNG_DR`.  I'll fix that now.  If you want something more sophisticated say what.

fnoble:
Ok, this makes more sense now

PR #397, opened 2015-05-06
Ephemeris debug

denniszollo:
Use the SBP definition of the ephemeris message

PR #486, opened 2015-07-07
Widen the PRN field

mfine:
Move the PRN field over to the SID field. https://github.com/swift-nav/libsbp/pull/185.

Incoming SID values are masked off with 0x1F where applicable (this is probably unnecessary as I think normal truncation will cover it - but just to be explicit) - outbound values are not masked. Note, this doesn't actually do any of the widening in the firmware, just supports widening around the messaging.

Are there other areas that are not covered here? Like, hand rolled messages?

/cc @henryhallam @gsmcmullin @fnoble

### DIFF
--- a/src/acq.c
+++ b/src/acq.c
@@ -48,7 +48,7 @@ void acq_send_result(u8 prn, float snr, float cp, float cf)
 {
   msg_acq_result_t acq_result_msg;
 
-  acq_result_msg.prn = prn;
+  acq_result_msg.sid = prn; /* TODO prn -> sid */
   acq_result_msg.snr = snr;
   acq_result_msg.cp = cp;
   acq_result_msg.cf = cf;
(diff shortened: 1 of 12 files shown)


mfine:
In particular, the prn/sid field in ephemeris message.

mfine:
Oh man, heading into libswiftnav to update ephemeris_t with the new sid field and the other new fields.

mfine:
Added pack/unpack routines from #397. Note, this message sizing will change (unpacked vs. packed, more messages), but hey, it's a new message id.

/cc @denniszollo

mfine:
This is ready for review.

mfine:
@mfine TODO: callback for the deprecated obs message to warn.

mfine:
Added in a deprecated message from old messages.

mfine:
HITL results look good for these changes - in line with good results for the release build - `v0.18-68-gcf`. Thoughts about merging this in after 19? @cbeighley @denniszollo
`

PR #487, opened 2015-07-07
Reduce false-positives in acquisition

soulcmdc:
Tied to the HDL change of the same name, but this treats the incoming average power as a fixed point number.  Unintended side effect of reducing the number of false positives in acquisition.

Includes @gsmcmullin stop-gap if SNR is NaN.  HDL now returns an obscenely large number for the average power if it's zero, so an error is now 

/cc @fnoble @gsmcmullin @cbeighley

### DIFF
--- a/src/acq.c
+++ b/src/acq.c
@@ -190,6 +190,11 @@ void acq_get_results(float* cp, float* cf, float* cn0)
   *cp = 1023.0 - (float)(acq_state.best_cp % (1023 * NAP_ACQ_CODE_PHASE_UNITS_PER_CHIP))
                   / NAP_ACQ_CODE_PHASE_UNITS_PER_CHIP;
   *cf = (float)acq_state.best_cf / NAP_ACQ_CARRIER_FREQ_UNITS_PER_HZ;
+  /* Only should occur if the the power returned from the HDL is exactly zero in any of the runs 
+   * Actual powers here are in the range of 5-10 */
+  if ((acq_state.power_acc / acq_state.count) > 100000) {
+    log_error("acq: Acquisition power out of range\n");
+  }
   /* "SNR" estimated by peak power over mean power. */
   float snr = (float)acq_state.best_power / (acq_state.power_acc / acq_state.count);
   if (snr == 0 || snr != snr) {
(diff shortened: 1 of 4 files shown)


soulcmdc:
HDL Link: https://github.com/swift-nav/Swift-NAP-HDL/pull/63

src/acq.c
```diff
@@ -102,7 +102,7 @@ static struct {
   u8 p_head;
   u8 p_tail;
 
-  u64 power_acc;      /**< Sum of powers of all acquisition set points. */
+  float power_acc;      /**< Sum of powers of all acquisition set points. */
```
fnoble: Indentation

src/acq.c
```diff
@@ -192,6 +192,11 @@ void acq_get_results(float* cp, float* cf, float* snr)
   *cf = (float)acq_state.best_cf / NAP_ACQ_CARRIER_FREQ_UNITS_PER_HZ;
   /* "SNR" estimated by peak power over mean power. */
   *snr = (float)acq_state.best_power / (acq_state.power_acc / acq_state.count);
+  if (isnan(*snr)) {
```
cbeighley: Maybe it doesn't matter, but maybe we should look for `power_acc == 0` directly?

src/acq.c
```diff
@@ -192,6 +192,12 @@ void acq_get_results(float* cp, float* cf, float* snr)
   *cf = (float)acq_state.best_cf / NAP_ACQ_CARRIER_FREQ_UNITS_PER_HZ;
   /* "SNR" estimated by peak power over mean power. */
   *snr = (float)acq_state.best_power / (acq_state.power_acc / acq_state.count);
+  if (acq_state.power_acc == 0) {
```
fnoble: Should we also catch the case where `acq_state.count` is zero?

fnoble:
Looks good, can you bump the required NAP version in `main.c` once the HDL release is cut?

src/acq.c
cbeighley: Can we add a comment explaining how/why this error conditions occurs?

cbeighley: OTOH maybe the text in `log_error` is explanatory enough.

soulcmdc:
@cbeighley : This is the other related commit.  @fnoble previously reviewed prior to rebase, I think this is just good to go.

soulcmdc:
Im melting

PR #488, opened 2015-07-08
Remove puts calls

mfine:
Quickie to pull out some leftover puts calls.

/cc @fnoble


fnoble:
These should eventually be changed to response messages, right?

mfine:
Right. There's a whole slew of messages like this that should potentially be moved over to error messages or related.

fnoble:
Happy to merge this as is for now. We can decide about responses later.

PR #464, opened 2015-06-30
[don't merge yet] PLL lock detect integration for HITL testing

gsmcmullin:

PR #489, opened 2015-07-08
PLL lock detector

henryhallam:
Based on @gsmcmullin's #464
Works OK on bench, ready for HITL


henryhallam:
@fnoble @gsmcmullin @soulcmdc Review would be appreciated while I continue to test and tune.

PR #499, opened 2015-07-13
Don't use a string setting for unknown serial number

### LINKED ISSUE #498: Use a different value for unknown serial id initialization
The unknown serial id causes problems in the console:

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/pyface/ui/qt4/gui.py", line 189, in _dispatch
    self._callable(*self._args, **self._kw)
  File "piksi_tools/console/console.py", line 270, in update_serial
    self.device_serial = 'PK%04d' % int(serial_string)
```

/cc @fnoble @henryhallam

henryhallam:
serial_number is -1 when those bytes in the NAP flash are unwritten.  So this does what we discussed, without any special-casing.
Fixes #498 
cc @mfine @fnoble

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -190,11 +190,7 @@ int main(void)
   sbp_fileio_setup();
   ext_setup();
 
-  if (serial_number < 0) {
-    READ_ONLY_PARAMETER("system_info", "serial_number", "(unknown)", TYPE_STRING);
-  } else {
-    READ_ONLY_PARAMETER("system_info", "serial_number", serial_number, TYPE_INT);
-  }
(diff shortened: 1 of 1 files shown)


mfine:
LGTM - not sure if it'll cause other problems elsewhere? @cbeighley

cbeighley:
Looks good - is TYPE_INT an s32?

mfine:
Yup:

```c
  static s32 serial_number;
```

henryhallam:
the settings macros / functions use sizeof to determine the length of the underlying variable.  So yes, in this case s32.

PR #500, opened 2015-07-15
Nmea overflow

henryhallam:
Was causing memory corruption.

### DIFF
--- a/src/nmea.c
+++ b/src/nmea.c
@@ -17,6 +17,7 @@
 
 #include <libswiftnav/coord_system.h>
 #include <libswiftnav/constants.h>
+#include <libswiftnav/logging.h>
 
 #include "board/nap/track_channel.h"
 #include "nmea.h"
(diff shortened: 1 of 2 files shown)


henryhallam:
@fnoble @gsmcmullin @cbeighley @denniszollo @ungureanuvladvictor What do you think of this approach? (minus the testing message)

src/nmea.c
```diff
@@ -78,27 +112,33 @@ void nmea_setup(void)
...
  */
-u8 nmea_checksum(char *s)
+static void nmea_append_checksum(char *s, size_t size)
 {
   u8 sum = 0;
+  char *p = s;
 
-  if (*s == '$')
-    s++;
+  if (*p == '$')
```
cbeighley: Add comment? i.e. `/* '$' header is not used in checksum calculation. */`
Also, maybe this should return an error if it doesn't start with `'$'`? It seems odd to support both cases - if it doesn't start with `'$'`, how do you know you're at the first character of the talker identifier immediately following `'$'`?

src/nmea.c
```diff
@@ -78,27 +112,33 @@ void nmea_setup(void)
...
+    p++;
 
-  while (*s != '*' && *s) {
-    sum ^= *s;
-    s++;
+  while (*p != '*' &&
```
cbeighley: Same? `/* '*' between data fields and checksum is not used in checksum calculation. */`

src/nmea.c
```diff
@@ -38,35 +39,68 @@ static struct nmea_dispatcher *nmea_dispatchers_head;
...
+ */
+#define NMEA_SENTENCE_START(max_len) char sentence_buf[max_len + NMEA_SUFFIX_LEN]; \
+  char *sentence_bufp = sentence_buf; \
+  char * const sentence_buf_end = sentence_buf + max_len;
+
+/** NMEA_SENTENCE_PRINTF: use like printf, can use multiple times
+    within a sentence. */
+#define NMEA_SENTENCE_PRINTF(fmt, ...) do { \
+    sentence_bufp += snprintf(sentence_bufp, sentence_buf_end - sentence_bufp, fmt, ##__VA_ARGS__); \
+    if (sentence_bufp >= sentence_buf_end) \
```
cbeighley: Should we send a `log_warn` if the NMEA sentence gets cut off?

src/nmea.c
henryhallam: I feel like it would just be code clutter to check for errors here (and then handle them in all the callers) - this is really just a helper function - but I'll add the comments.

Issue #65, opened 2014-07-29
Longer integration in tracking


fnoble:

soulcmdc:
@gsmcmullin  has implemented. @henryhallam working on parameters

PR #504, opened 2015-07-16
Track disabling state [don't merge]

henryhallam:
This doesn't seem to work well - the tracking channels seem to need a variable number of stop requests. 
@cbeighley


fnoble:
Doesn't fix #65, more work needed but closing for now

PR #508, opened 2015-07-17
Fix acq spins

henryhallam:
I think this should fix the occasional thing we've seen with the rapidly repeated SNR=0 messages, and maybe give us a little more insight into root cause.


henryhallam:
cc @cbeighley @denniszollo

cbeighley:
lgtm

PR #509, opened 2015-07-17
NMEA: GPVTG and GPRMC don't need nav measurements

henryhallam:
Just removing some unnecessary function parameters and calls.
cc @cbeighley @soulcmdc

### DIFF
--- a/src/nmea.c
+++ b/src/nmea.c
@@ -207,7 +207,7 @@ void nmea_gpgsa(const tracking_channel_t *chans, const dops_t *dops)
  * NMEA GPGSV message contains GPS satellites in view.
  *
  * \param n_used   Number of satellites currently being tracked.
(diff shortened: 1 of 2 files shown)


src/nmea.c
```diff
@@ -253,8 +253,7 @@ void nmea_gpgsv(u8 n_used, const navigation_measurement_t *nav_meas,
  * \param soln Pointer to gnss_solution struct
  * \param gps_t Pointer to the current GPS Time
  */
-void nmea_gprmc(const navigation_measurement_t *nav_meas,
-                const gnss_solution *soln, const gps_time_t *gps_t)
+void nmea_gprmc(const gnss_solution *soln, const gps_time_t *gps_t)
```
cbeighley: Docstring needs to be updated

cbeighley:
Otherwise looks good

PR #514, opened 2015-07-20
LIBSWIFTNAV: Calculate HDOP and VDOP correctly

henryhallam:
See swift-nav/libswiftnav#191


henryhallam:
Now with swift-nav/libswiftnav#198

PR #516, opened 2015-07-20
Use MSG_LOG

mfine:
Swift over to using MSG_LOG.

/cc @fnoble @gsmcmullin

### DIFF
--- a/src/sbp.c
+++ b/src/sbp.c
@@ -346,6 +346,9 @@ void log_(u8 level, const char *msg, ...)
   int n = vsnprintf(log->text, 255-sizeof(msg_log_t), msg, ap);
   va_end(ap);
 
+  if (n < 0)
+    return;
+
   sbp_send_msg(SBP_MSG_LOG, n+sizeof(msg_log_t), (u8 *)buf);
 }
 
--- a/src/sbp.c
+++ b/src/sbp.c
@@ -13,6 +13,7 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <stdio.h>
+#include <stdarg.h>
 #include <string.h>
 
 #include <ch.h>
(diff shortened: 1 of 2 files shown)


mfine:
Still needs a potential log mask setting to drive what levels get sent from the firmware. Not sure if we set our levels right:

```
#define LOG_DEBUG 0
#define LOG_INFO  1
#define LOG_WARN  2
#define LOG_ERROR 3
```

If we have a global setting (log_level), then it'd be everything and higher, if that's what we want. (e.g., set it to 3 to get only errors; 2 to get warns and errors; default of 0 gets debug and everything). Thoughts? @cbeighley @mookerji @denniszollo

gsmcmullin:
I'd suggest log levels used by syslog(3).  Lower numbers are higher priority.
```
#define LOG_EMERG       0       /* system is unusable */
#define LOG_ALERT       1       /* action must be taken immediately */
#define LOG_CRIT        2       /* critical conditions */
#define LOG_ERR         3       /* error conditions */
#define LOG_WARNING     4       /* warning conditions */
#define LOG_NOTICE      5       /* normal but significant condition */
#define LOG_INFO        6       /* informational */
#define LOG_DEBUG       7       /* debug-level messages */
```

mfine:
Here's another set of logging levels:

https://docs.python.org/2/library/logging.html#logging-levels

```
Level : Numeric value
CRITICAL : 50
ERROR : 40
WARNING : 30
INFO : 20
DEBUG : 10
NOTSET : 0
```

I've got no horse in this race - happy to do whichever scheme we're happy with.

src/sbp.c
```diff
@@ -331,6 +332,23 @@ int _write(int file, char *ptr, int len)
   }
 }
 
+/** Directs log_ output to the SBP log message */
+void log_(u8 level, const char *msg, ...)
+{
+  msg_log_t *log;
+  va_list ap;
+  char buf[256];
```
gsmcmullin: Use named constant for max payload size.

src/sbp.c
```diff
@@ -331,6 +332,23 @@ int _write(int file, char *ptr, int len)
...
+
+  log = (msg_log_t *)buf;
+  log->level = level;
+
+  va_start(ap, msg);
+  int n = vsnprintf(log->text, 255-sizeof(msg_log_t), msg, ap);
```
gsmcmullin: Use sizeof(buf) or named constant.

src/sbp.c
mfine: The named constant isn't in libsbp yet - will move over to the constant once it comes in.

mfine:
This is not incorporating a global `log level` setting - e.g., everything `<=` the global `log level` will be logged out from the firmware. What do we think about bringing something like that in?

PR #517, opened 2015-07-20
V0.19 branch off one

denniszollo:


fnoble:
Old and mysterious but it seems like most commits are in master... closing

PR #519, opened 2015-07-21
Work saturdays

henryhallam:
Prevents end-of-week failure to navigate; see swift-nav/libswiftnav#200

Tested on recorded RF from the weekend; handles both periods Sat 21:58:00...22:02:00 and Sat 23:58:00..Sun 00:02:00.

@denniszollo @cbeighley @mookerji @fnoble


henryhallam:
Need to test this in RTK as well as SPP, though I think it should work.

PR #522, opened 2015-07-21
check dimensionality properly before pvt or baseline solution

kovach:
should reduce the incidence of

"WARNING: dgnss_baseline returned error: -1"

and

"WARNING: PVT solver: Not enough measurements for solution (< 4) (code -7)"

### DIFF
--- a/src/solution.c
+++ b/src/solution.c
@@ -398,7 +398,7 @@ static msg_t solution_thread(void *arg)
     memcpy(nav_meas_old, nav_meas, sizeof(nav_meas));
     n_ready_old = n_ready;
 
-    if (n_ready_tdcp == 0) {
+    if (n_ready_tdcp < 4) {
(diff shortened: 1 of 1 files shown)


kovach:
cc @henryhallam @fnoble

src/solution.c
```diff
@@ -446,7 +446,9 @@ static msg_t solution_thread(void *arg)
                                     es, position_solution.time,
                                     sdiffs);
             chMtxUnlock();
-            output_baseline(num_sdiffs, sdiffs, &position_solution.time);
+            if (num_sdiffs >= 4) {
```
henryhallam: Confirm that 4 is the correct number of sdiffs for a baseline solution?

kovach: https://github.com/swift-nav/libswiftnav/blob/master/src/baseline.c#L578

PR #528, opened 2015-07-23
update flag output for raim changes to four sbp messages

kovach:


kovach:
cc @fnoble @henryhallam

src/sbp_utils.c
```diff
@@ -286,5 +286,26 @@ void pack_ephemeris(const ephemeris_t *e, msg_ephemeris_t *msg)
   msg->iode      = e->iode;
 }
 
+u8 sbp_format_baseline_flag(dgnss_baseline_t *solution)
+{
+  return solution->fixed_mode
+         | (solution->raim_available << 3)
```
henryhallam: Pity these bit offsets for the flag fields aren't exported in the libsbp generated C headers.  I suppose one could #define them here to avoid repeating magic numbers, but that would only be marginally prettier.

scottswift:
retest this please

PR #530, opened 2015-07-24
Changed simulation almanacs to const.

ungureanuvladvictor:
Freeing up ~3Kb. libswiftnav PR: https://github.com/swift-nav/libswiftnav/pull/202

/cc: @gsmcmullin @henryhallam

### DIFF
--- a/src/simulator_data.c
+++ b/src/simulator_data.c
@@ -11,7 +11,7 @@ u32 simulation_fake_carrier_bias[31];
 
 u8 simulation_num_almanacs = 31;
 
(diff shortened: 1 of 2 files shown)


gsmcmullin:
lgtm

henryhallam:
Good move!  Lately I've been tending to split piksi_firmware mods that require changes to libswiftnav into two separate commits, one that just changes the libswiftnav submodule and one that does everything else.  That can make rebases easier when there have been other libswiftnav changes.

PR #531, opened 2015-07-24
[Testing] Cleaning up stack sizes for diff threads

ungureanuvladvictor:
Do not merge

/cc: @fnoble @gsmcmullin @henryhallam 


ungureanuvladvictor:
num_sats: 11

(limit 6084)
obs_callback: 1620 5532

(limit 1000)
system_monitor_thread: 36 92

(limit 128)
track_status_thread: 12 96

(limit 22000)
time_matched_obs_thread: 1140 19168

(limit 12000)
solution_thread: 4900 7168

(limit 3000)
manage_track_thread: 8 178

(limit 3000)
manage_acq_thread: 4 480

ungureanuvladvictor:
With the simulator almanacs to flash we would have ~15kb free.

fnoble:
Looks good, although it is a risky change. ecstatic doesn't take into account stack usage of printf and some other functions so it isn't fool proof. That said, if its OK on HITL I'm fine to merge it.

ungureanuvladvictor:
First I ran it against ecstatic but after that I decided to watch the console and the stack usage reported there in a test environment (2 piksi hooked up with antennas on the roof and let them play together).

ungureanuvladvictor:
Tested on the simulator with 11 sats scenario for ~30 mins and looks ok.

PR #543, opened 2015-08-20
Try to match behavior of v0.20 with new LD + track mgmt arch

henryhallam:


denniszollo:
So this PR had noticeably worse performance in HITL testing than v0.20 are there more knobs to turn to try and get it to match?

gsmcmullin:
Included in #555

PR #545, opened 2015-08-24
Added checkpatch.pl for style checking.

ungureanuvladvictor:
It ignores spaces at the start of the line and not using tabs for indentation.

http://blog.vmsplice.net/2011/03/how-to-automatically-run-checkpatchpl.html to set up it as a githook on pre-commit (needs to run with --no-tree -f input.c).

\cc: @fnoble @kovach @imh 


fnoble:
Scott's work supersedes

PR #546, opened 2015-08-28
Uncrustify

kovach:
trying to satisfy checkpatch

haven't touched libswiftnav yet, wip


kovach:
TODO add back in (void) check

kovach:
checkpatch doesn't like the __DATE__ and __TIME__ macros

PR #547, opened 2015-08-28
Ensure setup.sh has pip on Linux

cbeighley:
@mookerji


cbeighley:
Version of arm tools also wasn't available in repo - @fnoble @gsmcmullin okay to change?

fnoble:
Should be fine. Did you try building the firmware with the new tools?

cbeighley:
Yes, it builds.

ungureanuvladvictor:
Compiler is good, I am using it for some time with the firmware.

PR #553, opened 2015-10-09
Added heading calculation and SBP logging.

benjamin0:
@fnoble please review, thanks!

### DIFF
--- a/src/solution.c
+++ b/src/solution.c
@@ -112,6 +112,15 @@ void solution_send_nmea(gnss_solution *soln, dops_t *dops,
 
 }
 
+double calc_heading(const double b_ned[3])
+{
+  double heading = atan2(b_ned[1], b_ned[0]);
+  if (heading < 0) {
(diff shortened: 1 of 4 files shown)


src/solution.c
```diff
@@ -112,6 +112,15 @@ void solution_send_nmea(gnss_solution *soln, dops_t *dops,
 
 }
 
+double calc_heading(const double b_ned[3])
+{
+  double heading = atan2(b_ned[1], b_ned[0]);
+  if (heading < 0) {
+    heading += 2 * M_PI;
+  }
+  return heading * (180 / M_PI);
```
fnoble: We have a constant `R2D` for this, nbd but makes it a little bit more readable.

https://github.com/swift-nav/libswiftnav/blob/master/include/libswiftnav/constants.h#L25

fnoble:
reviewed and looks great. Would be a good excuse to try interactive rebasing to squash down the multiple libsbp reference update commits.

PR #554, opened 2015-10-23
initial cut at only sending SPP when there is no psuedo absolute position

denniszollo:
This behavior will be determined by the "output best" setting. If output best is true, we will only send one kind of sbp/NMEA position messages.  if output_best if false, unit should behave as before.

### DIFF
--- a/src/solution.c
+++ b/src/solution.c
@@ -76,10 +76,7 @@ void solution_send_sbp(gnss_solution *soln, dops_t *dops, bool sent_psuedo_abs)
     /* Position in LLH. */
     /* if the device is configured to only output the best solution and
        we indicate that a psuedo absolute was sent, we do not output SPP*/
-    if (sent_psuedo_abs  && output_best) {
-      /* we do nothing in this case*/
-    }
-    else {
+    if (!(sent_psuedo_abs  && output_best)) {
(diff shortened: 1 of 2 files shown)


denniszollo:
@fnoble - can you look at this briefly?  I quickly set this up and am wondering what you think before doing anymore.  Basically I added a setting that says whether to continue outputting spp if we have psuedo absolute.

src/solution.c
```diff
@@ -193,7 +202,7 @@ void solution_send_baseline(const gps_time_t *t, u8 n_sats, double b_ecef[3],
   chMtxUnlock();
 }
 
-static void output_baseline(u8 num_sdiffs, const sdiff_t *sdiffs,
+static bool output_baseline(u8 num_sdiffs, const sdiff_t *sdiffs,
```
denniszollo: Added return codes to output_baseline

denniszollo:
@gsmcmullin -  can you look at this and see if you have a better idea?  I don't think mine is an elegant solution, but I would love to suppress SPP when we have a baseline and are operating in pseudo absolute mode.  It causes such confusion with customers and it necessitates a timeout to know when you've "lost" RTK output.

PR #557, opened 2015-11-03
Add PPS output to Piksi's debug connector

wltr:
Piksi generates a pulse-per-second synchronized with GPS time.

A thread periodically checks the current GPS time and time quality and
calculates the number of NAP clock cycles to the next full GPS second.
It then writes this value to an SPI register. Further, the pulse width
will be updated whenever the corresponding value is changed in Piksi
Console.

The PPS output has been tested in the lab with an oscilloscope using a
GPSDO as reference.


fnoble:
looks great, do you know what will happen if this is run against an older NAP version which doesn't have the PPS functionality?

wltr:
Just checked the FPGA code. If the firmware tries to write to an unknown SPI register, nothing will happen. The SPI registers check the first transmitted byte for their ID and if it doesn't match, they simply ignore it.

PR #558, opened 2015-11-03
Implement OpenLog peripheral driver

jacobmcnamee:
Try to detect OpenLog on UARTA/B at boot. If found, scan files until a unique name is found. Files are named according to <prefix><id>.bin, where <prefix> is a user-configurable string and <id> is auto-generated and unique.

Right now a new file will be started a) at boot, or b) when the prefix string is changed. There is also a function openlog_file_break() which will do the same. Eventually breaking the file (via function call or settings change) should probably be done in a separate background thread as it can potentially take seconds to find the next unique id.

### DIFF
--- a/src/peripherals/openlog.c
+++ b/src/peripherals/openlog.c
@@ -1,0 +1,591 @@
+/*
+ * Copyright (C) 2012-2015 Swift Navigation Inc.
+ * Contact: Jacob McNamee <jacob@swiftnav.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#include <string.h>
+#include <stdio.h>
+
+#include <ch.h>
+
+#include <libopencm3/stm32/f4/usart.h>
+
+#include <libswiftnav/logging.h>
+
+#include "openlog.h"
+#include "../settings.h"
+
+/** \addtogroup peripherals
+ * \{ */
+
+/** \defgroup openlog OpenLog
+ * Interface to the OpenLog open source SD data logger
+ *
+ * \{ */
+
+#define xstr(s) str(s)
+#define str(s) #s
+
+/** 3x ctrl-z */
+#define ESCAPE_SEQUENCE_STRING                "\x1A\x1A\x1A"
+/** string sent to execute a command */
+#define RETURN_STRING                         "\r\n"
+/** prompt returned on command success */
+#define COMMAND_PROMPT_STRING                 "\n>\n"
+/** prompt returned on command error */
+#define COMMAND_PROMPT_ERROR_STRING           "\n!>\n"
+/** prompt returned when entering data mode */
+#define DATA_PROMPT_STRING                    "<"
+#define DEFAULT_COMMAND_TERMINATORS           {COMMAND_PROMPT_STRING, \
+                                               COMMAND_PROMPT_ERROR_STRING}
+
+/* command strings */
+#define HELP_COMMAND_STRING                   "?"
+#define SIZE_COMMAND_STRING                   "size "
+#define APPEND_COMMAND_STRING                 "append "
+#define RESET_COMMAND_STRING                  "reset"
+
+/** string to search for in help output to detect OpenLog */
+#define HELP_RESPONSE_SEARCH_STRING           "OpenLog"
+
+/** also absorbs async transmit time */
+#define COMMAND_ECHO_TIMEOUT_ms               100
+#define RESPONSE_TIMEOUT_ms                   100
+#define HELP_RESPONSE_TIMEOUT_ms              500
+#define RECEIVE_DELAY_ms                      1
+#define CLAIM_TIMEOUT_ms                      1
+#define COMMAND_RETRY_COUNT                   3
+#define TERMINATORS_COUNT_MAX                 2
+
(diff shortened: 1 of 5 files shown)


fnoble:
Reviewing now - just a couple minor cosmetic points so far
 1. We use `/* */` style comments and not `//` in our C code
 2. Try to keep the max line length around 80 chars, going a bit over occasionally is fine but keeping lines around that length helps readability for some editor setups (and its in the style guide)

I'll follow up with any more substantive comments

src/peripherals/openlog.c
```diff
@@ -0,0 +1,524 @@
...
+static bool command_mode_enter(usart_dma_state *s, void *params); // enter/verify command mode
+static bool openlog_discover(usart_dma_state *s, void *params);
+static bool openlog_reset(usart_dma_state *s, void *params);
+static bool file_size_get(usart_dma_state *s, void *params);
+static bool file_append_set(usart_dma_state *s, void *params); // data mode entered on success
+
+static bool log_file_update();
+static bool file_prefix_setting_notify(struct setting *setting, const char *val);
+
+static bool claim(usart_dma_state *s)
```
fnoble: Could be moved up to `usart.c` and renamed e.g. `usart_claim_retry()`

src/peripherals/openlog.c
```diff
@@ -0,0 +1,524 @@
...
+      return true;
+    }
+
+    chThdSleepMilliseconds(CLAIM_FAIL_DELAY_ms);
+  } while (++retry < CLAIM_RETRY_COUNT);
+
+  return false;
+}
+
+static void release(usart_dma_state *s)
```
fnoble: Does this function need to exist?

src/peripherals/openlog.c
```diff
@@ -0,0 +1,524 @@
...
+{
+  usart_release(s);
+}
+
+static bool usart_send_str(usart_dma_state *s, const char *str)
```
fnoble: Again this is probably more generally useful, could move up to `usart.c`

src/peripherals/openlog.c
```diff
@@ -0,0 +1,524 @@
...
+  // null terminate
+  if (dest_index < dest_len) {
+    dest[dest_index++] = 0;
+    return true;
+  } else {
+    return false;
+  }
+}
+
+static bool parse_int(s32 *output, const char *str)
```
fnoble: what about using `atoi` or `strtoi` instead?

src/peripherals/openlog.c
```diff
@@ -0,0 +1,524 @@
...
+  bool success = false;
+
+  do {
+      success = func(s, params);
+  } while (!success && (++retry < COMMAND_RETRY_COUNT));
+
+  return success;
+}
+
+static bool strs_concat(char *dest, u32 dest_len, const char **srcs, u32 srcs_count)
```
fnoble: Use `strncat`? We have the full standard library, may as well use it :)

src/peripherals/usart.c
```diff
@@ -221,8 +228,16 @@ void usarts_enable(u32 ftdi_baud, u32 uarta_baud, u32 uartb_baud, bool do_precon
   chBSemInit(&uartb_state.claimed, FALSE);
   uartb_state.configured = true;
 
-  all_uarts_enabled = true;
+  if (do_configure_hooks) {
```
fnoble: Its not immediately obvious why these hooks aren't together with the radio ones. Maybe unify them so they are using the claim mechanism in the same way? or at least add a comment saying why they are separated.

src/peripherals/openlog.c
```diff
@@ -0,0 +1,524 @@
...
+/** Hook called when the UART is configured. Used to discover a connected OpenLog */
+void openlog_configure_hook(usart_dma_state *s, char* uart_name)
+{
+  if (openlog_usart != 0) {
+    // already discovered an OpenLog
+    return;
+  }
+
+  if (!claim(s))
+  {
```
fnoble: `{` should be on same line as `if`

src/peripherals/openlog.c
```diff
@@ -0,0 +1,524 @@
...
+    command_mode_enter(s, 0);
+    return false;
+  }
+}
+
+static bool log_file_update(usart_dma_state *s)
+{
+  if ((file_id < FILE_ID_START) || (file_id > FILE_ID_LAST))
+  {
+      return false;
```
fnoble: indentation and `{`

src/peripherals/openlog.c
```diff
@@ -0,0 +1,524 @@
...
+  }
+
+  // enter command mode
+  if (!command_execute(s, command_mode_enter, 0)) {
+    log_info("OpenLog - error entering command mode - resetting\n");
```
fnoble: Should some of these be `log_warn`s instead?

src/peripherals/openlog.c
```diff
@@ -0,0 +1,524 @@
...
+      break;
+    }
+    str_index++;
+  }
+
+  *output = (negative ? -1 : 1) * result;
+  return started;
+}
+
+static bool terminators_match(char c, const char **terminators, u32 *terminator_indexes, u32 terminators_count)
```
fnoble: I think this and/or `usart_wait_for_response` could do with a longer comment explaining the terminator mechanism.

fnoble:
This is great! I think a lot of people are going to find this really useful. Let me know when you have  addressed these comments and I'll do another pass.

src/peripherals/openlog.c
jacobmcnamee: added usart_claim_timeout() which accepts a semaphore timeout in systicks

src/peripherals/openlog.c
jacobmcnamee: They silently return zero on failure.

src/peripherals/usart.c
jacobmcnamee: The radio "preconfigure" hooks are a bit of a hack right now. They must be called before (and only before) any of the normal uart framework is operational. The openlog hooks use the standard claim() / release() mechanism. I think it's clear from the *pre*configure naming.

PR #561, opened 2015-11-10
Use signal_t from libswiftnav and libsbp.

gsmcmullin:


gsmcmullin:
retest this please

gsmcmullin:
retest this please

jacobmcnamee:
See inline comments. Mostly just var names.

src/manage.c
```diff
@@ -326,6 +336,7 @@ static void manage_acq()
   u8 prn = choose_prn();
   if (prn == (u8)-1)
     return;
+  gnss_signal_t sid = {.sat = prn};
```
jacobmcnamee: set constellation = CONSTELLATION_GPS and band = BAND_L1

gsmcmullin: I couldn't be bothered to fix this now.  In the next SBAS PR, `choose_prn` will become `choose_sid` and will return the sid with constellation and band filled in.

jacobmcnamee:
My only other concern is receiving SBP messages with sat >= 32 and causing an array overrun. Probably not a huge deal, but regardless it will be resolved when full signal_t support is implemented.

PR #564, opened 2015-11-16
Change AT commands for radio and UartA baud

rai-swiftnav:
Request to change the default 3DR radio settings. (Change the baud from 115 to 57 and air rate from 128 to 64) From bench testing, it was found that the 57600 baud worked much better that 115200. Along with the 3DR settings, this also changes the default uartA baud to work with the 3DR radios.

This change also requires to change settings.yaml in piksi_tools.

### DIFF
--- a/src/peripherals/3drradio.c
+++ b/src/peripherals/3drradio.c
@@ -32,7 +32,7 @@
 static const u32 baud_rates[] = {115200, 57600};//, 230400, 38400, 19200};
 
 /* This is the command string we send to the radios */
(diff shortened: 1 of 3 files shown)


rai-swiftnav:
Pull request for the Text change in Console
https://github.com/swift-nav/piksi_tools/pull/206

denniszollo:
@fnoble  Any thoughts on this change?  It may trip some users up, but I think is an improvement over all.

fnoble:
lgtm

denniszollo:
no longer relevant as was covered by #573

PR #565, opened 2015-11-16
Merge v0.20-branch into master

gsmcmullin:
This should have happened months ago.


denniszollo:
I don't think this needs to happen.  It is my understanding that all of these commits were back ported to the release branch from master except https://github.com/gsmcmullin/piksi_firmware/commit/7d08fcb2468c4fb236143301f6902b6cad0500f0

PR #566, opened 2015-11-17
Fix nmea bugs

denniszollo:
If you set one of the NMEA rates to 0, it would crash your Piksi.  Also, we should have a way to set message output rates to 0 if required.   

Changes: 
* Protect the DO_EVERY macro against a divide by zero
* Remove overwrite of user specified settings (that had a typo)

### DIFF
--- a/src/main.h
+++ b/src/main.h
@@ -22,7 +22,7 @@
  */
 #define DO_EVERY(n, cmd) do { \
   static u32 do_every_count = 0; \
-  if (if (n) > 0 && do_every_count % (n) == 0) { \
+  if ((n) > 0 && do_every_count % (n) == 0) { \
     cmd; \
   } \
   do_every_count++; \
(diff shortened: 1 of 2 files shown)


denniszollo:
@gsmcmullin@fnoble @soulcmdc.  Please take a look at this change, customer reported a crash when he set the some of the NMEA rates to zero.  

As a side note, we might also want to rename these settings to "period" or "interval".  People who use other receivers see the word "rate" and the start thinking the value is in hertz.

denniszollo:
My one possible problem is the silent failure when you set a bogus value for a setting, but we have a lot of those in the code thus far anyway.

gsmcmullin:
lgtm.
Sanity checking settings is a job for another day.
I agree with your suggested name change, but that is unrelated to the bug.

denniszollo:
OK merging with the future TODO of sanity checking settings and renaming the rate settings

PR #567, opened 2015-11-18
Use GPS time of reception in pseudorange calculation

fnoble:
See swift-nav/libswiftnav#233


ljbade:
Closing in favour of https://github.com/swift-nav/piksi_firmware/pull/627

PR #572, opened 2015-11-19
Generalize data structures PRN -> SID, SBAS support

jacobmcnamee:


src/track.c
```diff
@@ -84,17 +85,17 @@ static u16 iq_output_mask = 0;
...
 
-/* PRN lock counter
- * A map of PRN to an initially random number that increments each time that
- * PRN begins being tracked.
+/* signal lock counter
+ * A map of signal to an initially random number that increments each time that
+ * signal begins being tracked.
  */
-static u16 tracking_lock_counters[MAX_SATS];
+static u16 tracking_lock_counters[NUM_SATS];
```
gsmcmullin: This will need to change in future as other bands are supported, as the lock counters are per signal, not per satellite.

gsmcmullin:
lgtm.

ljbade:
@jacobmcnamee Excellent! I was waiting for this so I could start looking at other GNSS systems.

I am working on a branch to add QZSS L1C/A support. It only requires a few small differences from GPS.

I will push it once I get commit access.

PR #574, opened 2015-11-19
Reduce tracking state and thread state message rates

denniszollo:
This is an attempt to reduce serial throughput to improve perceived piksi performance when using the console.

Note that the 1hz rate of running the manage_track block (each second) should not have changed.


denniszollo:
@gsmcmullin I'm going to test this on the tablet PCs and see if it makes a difference

PR #581, opened 2015-12-07
MSG_BASE_POS_ECEF

mfine:
Support a `MSG_BASE_POS_ECEF` message, for both publishing and consuming - publish and consume base position in ECEF coordinates.

Handle receiving a base position in either coordinate system - similar behavior, no conversion necessary for ECEF.

Bring in new settings:
+ `surveyed_position->broadcast_ecef`
+ `surveyed_position->surveyed_ecef_x`
+ `surveyed_position->surveyed_ecef_y`
+ `surveyed_position->surveyed_ecef_z`

If `broadcast_ecef` is set, then publish the ECEF coordinates. Existing behavior for `broadcast` works unchanged. Another approach here might be to stop publishing LLH coordinates and phase out our consumption of them (existing `surveyed_position` settings would still be used, just published as ECEF).

https://github.com/swift-nav/libsbp/pull/320

/cc @fnoble @mookerji @denniszollo

### DIFF
--- a/src/base_obs.c
+++ b/src/base_obs.c
@@ -56,13 +56,13 @@ bool_t base_pos_known = false;
 double base_pos_ecef[3];
 
 /** SBP callback for when the base station sends us a message containing its
- * known location.
+ * known location in LLH coordinates.
  * Stores the base station position in the global #base_pos_ecef variable and
  * sets #base_pos_known to `true`.
  */
-static void base_pos_callback(u16 sender_id, u8 len, u8 msg[], void* context)
+static void base_pos_llh_callback(u16 sender_id, u8 len, u8 msg[], void* context)
 {
-  (void) context; (void) len; (void) msg; (void) sender_id;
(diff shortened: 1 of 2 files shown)


fnoble:
I'm not so sure this is a good idea. These new settings are confusing as most users don't have any clue about ECEF and it shouldn't be necessary to specify this separately. Also having two identical messages over the wire is going to be a pain, especially seeing as its more of an internal message rather than something that the user may want to consume in different formats (like the POS messages).

Happy to chat about this when you have a second.

mfine:
You're suggesting (a) unchanged settings and (b) publishing only in ECEF? That would be fine with me.

mfine:
@fnoble new firmware proposal: https://github.com/swift-nav/piksi_firmware/compare/master...mfine:mfine-msg-base-pos-ecef-2

mfine:
Switched over to the new proposal and squashed.

fnoble:
lgtm

PR #584, opened 2015-12-15
Add support for QZSS (WIP - do not merge)

ljbade:
Currently working on adding support for the Japanese QZSS constellation. To be used with https://github.com/swift-nav/libswiftnav/pull/253

It is designed to be compatible with GPS and only a few tweaks will be needed to support this system.

QZSS is visible in the Asia-Pacific region and will become a 7 satellite constellation with a mix of geostationary and  quasi-zenith orbits 

The main advantage of QZSS is the very high elevation over Australia and Japan, to counter limited sky visibility in high rise areas.

It also provides many additional services and signals to boost accuracy in Japan. Finally QZSS is the only on-orbit implementation of the modernised GPS L5 and L1C signals.


ljbade:
Tested the branch this afternoon, Piksi is now tracking PRN 193 (but for some reason it reads as 194 in the console)

image

However after a few seconds it dies:
`One or more threads appear to be dead: 0x0000000E. Watchdog reset imminent.`

ljbade:
I fixed the watchdog reset, and everything seems to be working. PRN number is still off though.

ljbade:
Fixed the PRN number. Seems 1 is added to the numbers in `signal.h`.

Also did a RTK test and after several minutes I got an RTK fixed solution.

PR #600, opened 2016-01-16
Obs period monitor

denniszollo:
This provides monitoring information of how often observations are being received via a new UartStateMessage.
cc @fnoble  per our discussion

### DIFF
--- a/src/base_obs.c
+++ b/src/base_obs.c
@@ -255,12 +255,6 @@ static void obs_callback(u16 sender_id, u8 len, u8 msg[], void* context)
     return;
   }
 
-  /* Calculate packet latency. */
-  if (time_quality >= TIME_COARSE) {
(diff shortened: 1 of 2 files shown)


denniszollo:
/cc @jacobmcnamee  @benjamin0  any time for code review of this?  

This PR will allow us to see through the logs how often base station observations come in to assess radio link and skylark link reliability.  Since we don't forward observations from the base station out of the non USB UARTS, we don't have a clue how well observations are coming through right now when we listen on UARTB.  

It also requires the following commits of tools: https://github.com/swift-nav/piksi_tools/pull/239 and sbp: https://github.com/swift-nav/libsbp/pull/329

After review I will do some HITL testing and will update sbp_log_analysis for the new message (but I don't anticipate any change from HITL testing).

PR #602, opened 2016-01-21
remove cw channel NAP hooks

denniszollo:
/cc @cbeighley  - thoughts on taking the CW out of firmware source?


denniszollo:
Any problems with this?  Since we don't have CW filtering support in the NAP, I removed it from the firmware for the time-being.

/cc @fnoble @jacobmcnamee @cbeighley @gsmcmullin

gsmcmullin:
lgtm.  I'm always in favour of removing dead code.  If this does come back it will most likely be redone to use the FFT core.

denniszollo:
With Gareth's feedback (and crickets from everyone else) I merged this.

PR #604, opened 2016-01-26
Enable WFI idle

jacobmcnamee:
Enables execution of WFI power-saving instruction from the idle thread. Should reduce CPU power consumption significantly.


gsmcmullin:
lgtm. Merge at will.

PR #588, opened 2015-12-17
Fixes to use new ephemeris functions

ljbade:
Needed for https://github.com/swift-nav/libswiftnav/pull/255

PR #606, opened 2016-01-29
L2C capability monitor

dt-exafore:
https://github.com/swift-nav/exafore_planning/issues/38

### DIFF
--- a/src/l2c_capability.c
+++ b/src/l2c_capability.c
@@ -1,0 +1,133 @@
+/*
+ * Copyright (C) 2016 Swift Navigation Inc.
+ * Contact: Dmitry Tatarinov <dmitry.tatarinov@exafore.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+#include <stdio.h>
+#include <ch.h>
+#include <assert.h>
+
+#include <libswiftnav/common.h>
+#include <libswiftnav/nav_msg.h>
+#include <libswiftnav/logging.h>
+
+#include "cfs/cfs-coffee.h"
+#include "cfs/cfs.h"
+
+#include "l2c_capability.h"
+
+static MUTEX_DECL(l2c_capability_upd_mutex);
+
+/** gps_l2c_sv_capability 
+  L2C capability info containter
+  Each bit responses for a SV:
+  Bit 0 - SV 1, Bit 31 - SV 32.
+  bit set - L2C stream available for the SV
+  bit cleared - L2C stream not available
+  By default assume all SV can broadcast L2C stream */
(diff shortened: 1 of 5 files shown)


dt-exafore:
@adel-mamin @ljbade @jacobmcnamee Please, review.
NOTE: to pass all checks following PRs should be merged: 
https://github.com/swift-nav/piksi_firmware/pull/588
https://github.com/swift-nav/libswiftnav/pull/294

dt-exafore:
@adel-mamin @ljbade @jacobmcnamee Please, review.
NOTE: to pass all checks following PRs should be merged: 
#588
swift-nav/libswiftnav#294

src/l2c_capability.c
```diff
@@ -0,0 +1,128 @@
...
+
+/** Return current value of gps_l2c_sv_capability
+ *  \param u32 output value of L2C capability
+ *
+ * \return L2C capabapility
+ */
+void l2c_capability_get(u32 *l2c_cpbl)
+{
+  chMtxLock(&l2c_capability_upd_mutex);
+  *l2c_cpbl = gps_l2c_sv_capability;
```
adel-mamin: Add check l2c_cpbl against NULL

src/l2c_capability.c
```diff
@@ -0,0 +1,134 @@
...
+  By default assume all SV can broadcast L2C stream */
+static u32 gps_l2c_sv_capability = 0xffffffff;
+
+/** \defgroup l2c_capability L2C_Capability
+ * Functions used in l2c capability
+ * \{ */
+
+/** Update gps_l2c_sv_capability. And write it to Flash memory if
+ * previous value of gps_l2c_sv_capability is different.
+ * \param u32 New value of L2C capability
```
adel-mamin: u32 -> l2c_cpbl

src/l2c_capability.c
```diff
@@ -0,0 +1,134 @@
...
+    cfs_close(fd);
+  } else
+    log_error("Error opening L2C capability file");
+
+  Mutex *m = chMtxUnlock();
+  assert(m == &l2c_capability_upd_mutex);
+}
+
+/** Return current value of gps_l2c_sv_capability
+ *  \param u32 output value of L2C capability
```
adel-mamin: u32 -> l2_cpbl

src/l2c_capability.c
```diff
@@ -0,0 +1,133 @@
...
+  if (l2c_cpbl == gps_l2c_sv_capability)
+    return;
+
+  chMtxLock(&l2c_capability_upd_mutex);
+
+  gps_l2c_sv_capability = l2c_cpbl;
+
+  int fd = cfs_open(L2C_FILE, CFS_WRITE);
+  if (fd != -1) {
+    if(cfs_write(fd,
```
jacobmcnamee: Flash writes from the decoder thread scare me a little bit. I don't know what the maximum write time is, but if we do this for several channels back to back it might cause the nav bit FIFOs to overrun. Right now the FIFOs are 32 nav bits per channel, which can be as little as 64ms for some encodings.

I think doing the writes directly can be ok as long as we're sure timing won't be violated, but a better option might be to set a flag and periodically write to flash from a background thread when necessary (assuming the write can be done in a way that doesn't block the decode thread).

dt-exafore: Since all SVs send SV config data for whole constellation incoming L2C data is same from all channels. Current algorithm monitors changing L2C capability data, so if we update it from one channel it will not updated from others. Worst case is flash writing rate is once per 12.5 mins.

dt-exafore:
https://github.com/swift-nav/piksi_firmware/pull/630 should be merged first.

PR #607, opened 2016-02-02
Remove libopencm3 dependency.  Use ChibiOS/HAL.

gsmcmullin:
First cut at replacing libopencm3 and porting to ChibiOS/HAL.  Sorry for one giant commit, but doing it incrementally leaves conflicts between ChibiOS/HAL and libopencm3.

Use ChibiOS/HAL PAL, EXT, SPI, Serial and GPT driver abstractions for relevant peripherals.
- SPI driver with DMA doesn't work: results in DMA errors (disabled in this PR).
- SPI code is wrapped for chip select handling.
- Serial driver is interrupt driven (no DMA)
- UART code is wrapped for configuration and statistics gathering.
- 3DR radio code reworked to use `usart_chat.h` functions.

libopencm3 dependency removed.
- RNG functions use ChibiOS provided STM32 definitions (abstraction is OK).
- WDT functions use ChibiOS provided STM32 definitions (abstraction is OK, newer ChibiOS/HAL provides WDG driver interface).
- Flash routines directly implemented in piksi_firmware (abstraction is OK).
- Reset reason, fallback usart (on error) and fault handling need better abstraction, these do register level access.

/cc @jacobmcnamee


src/board/leds.c
```diff
@@ -26,8 +25,8 @@
 /** Setup the LED GPIOs. */
 void led_setup(void)
 {
-  RCC_AHB1ENR |= RCC_AHB1ENR_IOPCEN;
-  gpio_mode_setup(GPIOC, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, GPIO3 | GPIO4);
+  palSetPadMode(GPIOC, GPIOC_LED1, PAL_MODE_OUTPUT_PUSHPULL);
```
jacobmcnamee: Not sure if it's in Chibi 2, but in 3 there is a GPIO line type which combines port and pad.

gsmcmullin: Ok.  I agree this is pretty janky.  I'm busy with the port to ChibiOS 3 now.  If that will clean it up we can wait for those changes.  Actually this code can go, as all the ports/pads are initialized in halInit() anyway, to the values in board.h.  It's probably cleaner to have it only done in one place.

src/solution.c
```diff
@@ -286,39 +293,35 @@ void send_observations(u8 n, gps_time_t *t, navigation_measurement_t *m)
...
 
-static void timer_set_period_check(uint32_t timer_peripheral, uint32_t period)
+static void timer_set_period_check(GPTDriver *gptd, uint32_t period)
 {
-  __asm__("CPSID i;");
-  TIM_ARR(timer_peripheral) = period;
-  uint32_t tmp = TIM_CNT(timer_peripheral);
+  chSysLock();
+  gptChangeIntervalI(gptd, period);
+  u32 tmp = gpt_lld_get_counter(gptd);
```
jacobmcnamee: gptGetCounterX()

gsmcmullin:
Replaced with #612

PR #612, opened 2016-02-09
Port to Chibios 3 and ChibiOS/HAL.

gsmcmullin:
Update ChibiOS kernel interface to ChibiOS/RT 3.1.  Use ChibiOS/HAL PAL, EXT, SPI, Serial, GPT and  WDT driver abstractions for their respective peripherals.
- libopencm3 dependency removed.
- SPI driver with DMA doesn't work: results in DMA errors (disabled in this PR).
- SPI code is wrapped for chip select handling.
- Serial driver is interrupt driven (no DMA)
- UART code is wrapped for configuration and statistics gathering.
- 3DR radio code reworked to use usart_chat.h functions.
- Watchdog code still include STM32 specific stuff to set the period.
- RNG functions use ChibiOS provided STM32 definitions (abstraction is OK).
- Flash routines directly implemented in piksi_firmware (abstraction is OK).
- Reset reason, fallback usart (on error) and fault handling need better abstraction, these do register level access.


src/solution.c
```diff
@@ -285,40 +295,35 @@ void send_observations(u8 n, gps_time_t *t, navigation_measurement_t *m)
...
 {
-  __asm__("CPSID i;");
-  TIM_ARR(timer_peripheral) = period;
-  uint32_t tmp = TIM_CNT(timer_peripheral);
+  chSysLock();
+  gptChangeIntervalI(gptd, period);
+  u32 tmp = gptGetCounterX(gptd);
   if (tmp > period) {
-    TIM_CNT(timer_peripheral) = period;
+    gpt_lld_set_counter(gptd, period);
```
jacobmcnamee: This won't work on Zynq. There is no way to modify the counter.

src/peripherals/watchdog.c
```diff
@@ -28,15 +46,45 @@
...
+  } else {
+  	exponent = IWDG_PR_DIV4;   reload = (count >> 2);
+  }
+
+  /* Avoid the undefined situation of a zero count */
+  if (count == 0) {
+  	count = 1;
+  }
+
+  wdg_config.pr = exponent;
```
jacobmcnamee: Would it be better / easier to #define the whole watchdog config in board.h?

src/solution.c
```diff
@@ -38,10 +40,18 @@
 #include "timing.h"
 #include "base_obs.h"
 #include "ephemeris.h"
-#include "./system_monitor.h"
+#include "system_monitor.h"
+
+#define SOL_GPT GPTD5
+
+static void sol_gpt_cb(GPTDriver *);
+static GPTConfig sol_gpt_config = {
```
jacobmcnamee: Same here - probably want to define all the driver configs in board.h

src/solution.c
gsmcmullin: What happens if the interval is set to a value less than the current counter?  This check is meant to catch that condition.  This code is not portable anyway, the meaning of the counter value is explicitly undefined.

jacobmcnamee: That's exactly why my Zynq GPT driver does this:

```void gpt_lld_change_interval(GPTDriver *gptp, gptcnt_t interval) {

  /* If the new interval value happens to be less than the current counter
   * value, then the counter will not reset until it overflows. To prevent
   * this, simply stop and restart the timer with the new interval.
   */

  gpt_lld_stop_timer(gptp);
  gpt_lld_start_timer(gptp, interval);
}```

PR #613, opened 2016-02-14
Fix bug in decoder_interface_register()

jacobmcnamee:


src/decode.c
```diff
@@ -125,7 +125,7 @@ void decoder_interface_register(decoder_interface_list_element_t *element)
   decoder_interface_list_element_t **e = &decoder_interface_list;
 
   while (*e != 0)
-    *e = (*e)->next;
+    e = &(*e)->next;
```
ljbade: Not sure how it could be changed but this function is fairly hard to follow with the double pointer

PR #614, opened 2016-02-15
Support signal_t refactor, add platform-specific code support

jacobmcnamee:
Adds defines and indexing functions enabling platform-specific support of individual code types.
```
/* Platform-specific code support ==== */
#define CODE_GPS_L1CA_SUPPORT     1
#define CODE_GPS_L2CM_SUPPORT     0
#define CODE_SBAS_L1CA_SUPPORT    1
/* =================================== */
```
Unsupported codes are excluded from signal count defines and indexing functions.

Support for swift-nav/libswiftnav#303.

### DIFF
--- a/src/signal.c
+++ b/src/signal.c
@@ -1,0 +1,164 @@
+/*
+ * Copyright (C) 2011-2016 Swift Navigation Inc.
+ * Contact: Jacob McNamee <jacob@swiftnav.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#include "signal.h"
+
+#include <assert.h>
+#include <string.h>
+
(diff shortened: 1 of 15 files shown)


jacobmcnamee:
Requires update to libswiftnav pending PR.

.gitmodules
```diff
@@ -3,7 +3,7 @@
 	url = https://github.com/libopencm3/libopencm3.git
 [submodule "libswiftnav"]
 	path = libswiftnav
-	url = https://github.com/swift-nav/libswiftnav.git
+	url = https://github.com/jacobmcnamee/libswiftnav.git
```
pmiettinen: You probably don't mean to change this submodule url?

jacobmcnamee: No, it's just for testing until the libswiftnav change is merged.

src/signal.c
```diff
@@ -0,0 +1,164 @@
+/*
+ * Copyright (C) 2011-2016 Swift Navigation Inc.
```
pmiettinen: Should be only '2016'?

src/signal.h
```diff
@@ -0,0 +1,58 @@
...
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef SWIFTNAV_SIGNAL_H
+#define SWIFTNAV_SIGNAL_H
+
+#include <libswiftnav/common.h>
+#include <libswiftnav/signal.h>
+
+/* Platform-specific code support ==== */
```
pmiettinen: Is the plan to ultimately set these defines in build system?

jacobmcnamee: Yes, definitely. Either makefile or board.h. Just waiting for other changes to set this up.

PR #615, opened 2016-02-16
Check the alert flag from nav message

ljbade:
Requires https://github.com/swift-nav/libswiftnav/pull/304 and https://github.com/swift-nav/libswiftnav/pull/316

### DIFF
--- a/src/decode_gps_l1.c
+++ b/src/decode_gps_l1.c
@@ -94,6 +94,14 @@ static void decoder_gps_l1_process(const decoder_channel_info_t *channel_info,
         log_warn_sid(channel_info->sid, "TOW set failed");
       }
     }
+
+    /* Update alert flag */
+    u8 alert = data->nav_msg.alert;
+    if (!tracking_channel_alert_sync(channel_info->tracking_channel, alert)) {
+      char buf[SID_STR_LEN_MAX];
+      sid_to_string(buf, sizeof(buf), channel_info->sid);
+      log_warn("%s alert set failed", buf);
+    }
   }
 
   /* Check if there is a new nav msg subframe to process. */
--- a/src/decode_gps_l1.c
+++ b/src/decode_gps_l1.c
@@ -94,14 +94,6 @@ static void decoder_gps_l1_process(const decoder_channel_info_t *channel_info,
         log_warn_sid(channel_info->sid, "TOW set failed");
       }
     }
-
-    /* Update alert flag */
-    u8 alert = data->nav_msg.alert;
-    if (!tracking_channel_alert_sync(channel_info->tracking_channel, alert)) {
-      char buf[SID_STR_LEN_MAX];
-      sid_to_string(buf, sizeof(buf), channel_info->sid);
-      log_warn("%s alert set failed", buf);
-    }
   }
 
   /* Check if there is a new nav msg subframe to process. */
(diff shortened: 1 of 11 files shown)


ljbade:
@swift-nav/firmware Ready for review

Running HITL tests

ljbade:
Rebased, need to test locally and on HITL.

ljbade:
From http://waas.stanford.edu/papers/Walter_IONPNT_2015_ARAIM_characterization.pdf

Should also check for repeated nav data parity failures, then stop tracking the sat and delete the ephemeris data.

After satellite comes back from unhealthy, need to be careful to wait for new ephemeris and not rely on old data that might still say "healthy".

ljbade:
HITL test seems fine, but number of warnings was high due to an unhealthy satellite.

Should an unhealthy satellite be a warning or info level message?

ljbade:
More improvements: Use the valid ranges of the various ephemeris parameters to detect incorrectly decoded ephemeris and skip it.

See http://www.gps.gov/technical/icwg/meetings/2015/PIRN-IS-200H-003.pdf

ljbade:
Going to bring in the additional validation from https://github.com/swift-nav/libswiftnav/pull/316

ljbade:
Implemented the changes. Need to test.

ljbade:
Should check how to properly handle the Integrity Status Flag

ljbade:
Need to be able to detect these:
from http://gps.stanford.edu/papers/Walter_IONITM_2016_Ephemeris.pdf

“Alert – Alarm Indications. An
otherwise healthy SPS SIS or marginal SPS SIS becomes
unhealthy when it is the subject of a SPS SIS alarm
indication. The presence of any one of the 9 alarm
indications listed below means the information provided
by the SPS SIS may not be correct. The SPS SIS alarm
indications are defined to include the following:
(1) The SPS SIS becomes untrackable (e.g., ≥ 20 dB
decrease in transmitted signal power, ≥ 20 dB increase
in correlation loss):
(a) The SPS SIS ceases transmission.
(b) The elimination of the standard C/A-code.
(c) The substitution of non-standard C/A-code for the
standard C/A-code.
(d) The substitution of PRN C/A-code number 37 for
the standard C/A-code.
(2) The failure of parity on 5 successive words of NAV
data (3 seconds).
(3) The broadcast IODE does not match the 8 LSBs of
the broadcast Index of Data Clock (IODC) (excluding
normal data set cutovers, see IS-GPS-200).
(4) The transmitted bits in subframe 1, 2, or 3 are all
set to 0's or all set to 1's.
(5) Default NAV data is being transmitted in subframes
1, 2, or 3 (see IS-GPS-200).
(6) The 8-bit preamble does not equal 10001011,
decimal 139, or hexadecimal 8B.”

src/decode_gps_l1.c
```diff
@@ -102,10 +102,26 @@ static void decoder_gps_l1_process(const decoder_channel_info_t *channel_info,
...
 
-  if (ret <= 0)
     return;
+  }
+
+  /* Update alert flag */
+  u8 alert = data->nav_msg.alert;
+  alert |= data->nav_msg.as;
+  if (!tracking_channel_alert_sync(channel_info->tracking_channel, alert)) {
+    char buf[SID_STR_LEN_MAX];
```
jacobmcnamee: Use `log_warn_sid()`

src/ephemeris.c
```diff
@@ -99,6 +99,21 @@ void ephemeris_new(ephemeris_t *e)
...
 }
 
+void ephemeris_clear(gnss_signal_t sid) {
+  assert(sid_supported(sid));
+
+  u32 index = sid_to_global_index(sid);
+
+  char buf[SID_STR_LEN_MAX];
+  sid_to_string(buf, sizeof(buf), sid);
+  log_info("Clearing ephemeris for %s", buf);
```
jacobmcnamee: `log_info_sid()`

jacobmcnamee:
I think we should put the alert flag status in the NDB (https://swiftnav.hackpad.com/Piksi-Nav-DB-SMUivUFDh0Q#:h=Information-Elements) instead of in the tracking channels. Probably much cleaner that way.

ljbade:
@jacobmcnamee How often is the NDB updated? Alert flag can change every 0.6 seconds (one Word)

jacobmcnamee:
It's a general datastore. It can be updated as often as necessary.

ljbade:
@jacobmcnamee Ah OK, I will look into using it then

ljbade:
@jacobmcnamee What is the status of the nav DB? Did it get merged?

I would like to finish off support for the alert flag.

PR #617, opened 2016-02-19
Ignore Eclipse IDE-specific files

rge-exafore:
This commit adds Eclipse IDE-specific files to .gitignore file
so git is instructed not to track changes in those files.


ljbade:
Looks good

PR #621, opened 2016-02-21
WIP: Add Nav DB module

rge-exafore:
This change adds Nav DB module.

Issues: 
https://github.com/swift-nav/exafore_planning/issues/55
https://github.com/swift-nav/exafore_planning/issues/63

Design doc:
https://swiftnav.hackpad.com/Piksi-Nav-DB-SMUivUFDh0Q


ljbade:
Nice, I like the design. It will be very easy for me to expand to add support for ionosphere and UTC parameters.

src/ndb.c
```diff
@@ -0,0 +1,121 @@
...
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#include <string.h>
+#include "ndb.h"
+
+#include "ndb_p1_section.h"
+#include "ndb_p3_section.h"
+
+ndb_file_version_t ndb_file_version = 0x00000001;
```
jacobmcnamee: I think we should just use the firmware version.

src/ndb.c
```diff
@@ -0,0 +1,121 @@
...
+
+#define NDB_MAX_SECTIONS_NUM 2
+static ndb_section_if_t *ndb_sections[NDB_MAX_SECTIONS_NUM];
```
jacobmcnamee: Add a next pointer to ndb_section_if_t and make this a linked list?

src/ndb_common.h
```diff
@@ -0,0 +1,61 @@
...
+    NDB_ERR_SECTION_INIT_FAILURE,
+    NDB_ERR_INIT_DONE
+};
+
+enum ndb_data_source
+{
+    NDB_DS_UNDEFINED, NDB_DS_RECEIVER, NDB_DS_SBP
+};
+
+enum ndb_information_element
```
jacobmcnamee: I'd much rather have dedicated functions for each data element defined in the respective header files. Something like
```
ephemeris_t ephemeris;
if(ndb_gps_l1ca_eph_read(sid, &ephemeris)) {
  /* Do something */
}
```
would be much more user friendly and hard to mess up than
```
ephemeris_t ephemeris = {.sid = sid};
ndb_request_info_t request_info = {
  .ie= NDB_IE_GPS_L1CA_EPHEMERIS,
  .data = (ndb_ie_t){.ephemeris = &ephemeris}
};
if(ndb_retrieve_ie(&request_info)) {
  /* Do something */
}
```

src/ndb_core.c
```diff
@@ -0,0 +1,245 @@
...
+
+/** Write request */
+typedef struct
+{
+    ndb_file_t *file;
+    ndb_element_t element;
+} ndb_write_q_node_t;
+
+/* TODO: Move to platform parameterization module */
+#define NDB_WRITE_QUEUE_SIZE 8
```
jacobmcnamee: I haven't thought about this much, but would it be better to add a `nvm_dirty` flag to the metadata struct for each element and then periodically scan through all the elements in the nvm write thread?

src/ndb_p1_section.c
```diff
@@ -0,0 +1,183 @@
...
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+
+#include "ndb_p1_section.h"
+
+#include <string.h>
+#include "ndb.h"
+
+#define GPS_L1_EPHE_FILE_NAME "gps_l1_ephe"
```
jacobmcnamee: Another design question: Would it be better / easier to NOT split out data elements that are common across all codes? E.g. why not just do ephemeris_t ephemeris[PLATFORM_SIGNAL_COUNT] and then have a generic ephemeris_get(gnss_signal_t sid, ephemeris_t *ephemeris) function?

src/ndb.c
```diff
@@ -0,0 +1,121 @@
...
+                ndb_sections[i] = section_if;
+                return NDB_ERR_NONE;
+            }
+            return NDB_ERR_SECTION_INIT_FAILURE;
+        }
+    }
+    return NDB_ERR_NO_SECTION_SPACE;
+}
+
+enum ndb_op_code ndb_process_req(ndb_request_info_t *r, size_t f_offset)
```
jacobmcnamee: I think the core module needs to be aware of (or at least be able to find) the registered data elements for purposed of writing to NVM, but I don't really see a reason why the read/write functions need to pass from core -> section-specific -> core. Why not just go straight from section-specific -> core?

src/ndb.c
rge-exafore: I agree. Where is it defined?

src/ndb.c
rge-exafore: Yes, sure. Done.

src/ndb_common.h
rge-exafore: My plan is to have wrapper functions like your first example around update/retrieve function calls like in your second example. However, I'd like to keep just two API points to the DB itself: update and retrieve. I just follow the classic architecture here: usually we have SELECT and UPDATE and build some logic on top of that. If you think we should switch to the approach you proposed I'll change the design.

src/ndb_p1_section.c
rge-exafore: I guess it's a matter of preference. I cannot mathematically prove that but my intuition says that keeping things separate makes sense. However, I prepared a version of code with NDB_COMMON_STORAGE #ifdef (temporary one, I'll delete it). Please review and let me know which branch to delete.

src/ndb_core.c
rge-exafore: Changed.

src/ndb.c
rge-exafore: I didn't quite understand that. Sections don't have public API. That's why it goes Public API-> section -> core. Am I missing something?

src/ndb.c
jacobmcnamee: `GIT_VERSION` is #defined to a string like `v0.21-77-g33f9988-dirty` for non-release builds and something like `v0.21` for release builds.

I doubt you need it, but there is also a read-only setting which holds the version:
`READ_ONLY_PARAMETER("system_info", "firmware_version", GIT_VERSION, TYPE_STRING);`

src/ndb.c
```diff
@@ -0,0 +1,138 @@
...
+    ndb_register_section(&p3_section_if);
+
+    return NDB_ERR_NONE;
+}
+
+void ndb_register_section(ndb_section_if_t* section_if)
+{
+    ndb_section_if_t **p = &ndb_first_section;
+    while(NULL != *p){
+        *p = (*p)->next_section;
```
jacobmcnamee: I think you want `p = &(*p)->next_section`. I just fixed fixed the decoder https://github.com/swift-nav/piksi_firmware/commit/6d22acecf12b61284c399e0e0518bf7e7d5b47d7.

src/ndb_common.h
jacobmcnamee: I don't see why the core needs to be aware of the different data types, though. If the core only operated on generic `ndb_element_t` types, then you could put the public API functions in the individual sections and have those directly call the core functions.

Looking at `ndb_process_req()`, it seems like the only benefit is returning an error if a data element is not supported on the current hardware platform. I.e. the second code snippet would compile and run but would return `NDB_ERR_UNSUPPORTED_FUNCTION`, whereas the first snipped would not even be compiled/linked.

src/ndb_p1_section.c
jacobmcnamee: I think in general we should combine where possible. Although specifically for almanac and ephemeris, those are probably one per physical satellite and not one per signal. Maybe I should add an `sid_to_satellite_index()` API or something.

src/ndb_common.h
jacobmcnamee: Ok, I am reasonably confident my second paragraph above is accurate: The benefit of the more complicated approach is that all the NDB functions exist on all hardware platforms, but unsupported data would yield `NDB_ERR_UNSUPPORTED_FUNCTION`. I do think this will be useful in a handful of places (e.g. L2C capability), but there are easier and more performant ways of getting the same functionality. Worst case, we could manually implement or #define unsupported functions to return an error. Better, we could add something like `__attribute__ ((weak, alias("ndb_unimplemented")))` to all the public NDB API functions. That way anything that isn't resolved by the linker will be aliased to a function which returns `NDB_ERR_UNSUPPORTED_FUNCTION`.

src/ndb.c
```diff
@@ -0,0 +1,215 @@
...
+
+static ndb_thread_t ndb_thread;
+static WORKING_AREA(ndb_thread_wa, 128);
+static msg_t nv_write_thread(void*);
+static ndb_mutex_t data_access;
+
+static enum ndb_op_code ndb_open_file(ndb_file_t *file, ndb_file_version_t *v);
+static enum ndb_op_code ndb_read(ndb_file_t *f, void *first_element, size_t s);
+
+#define OPTF __attribute__ ((weak, alias ("ndb_not_implemented")))
```
jacobmcnamee: FYI for the public APIs, there are a couple intricacies involved in getting this to work.
1.  `ndb_not_implemented()` needs to itself be *defined* as a weak symbol `__attribute__ ((weak))` in the header files where `__attribute__ ((weak, alias ("ndb_not_implemented")))` is applied to other function prototypes. I.e. you need something like 
```bool __attribute__ ((weak)) fake_unimplemented(void)
{
  return false;
}```
in the header file (probably a common header included by all the public API headers).
2. Said function prototypes with `__attribute__ ((weak, alias ("ndb_not_implemented")))` should NOT be visible to the source file where the actual implementations appear (otherwise you'll get redefinition errors).

File: src/ndb_p1.c [hunk 9]
```diff
@@ -0,0 +1,134 @@
...
+
+    if(ndb_load_data(&ndb_alma_file, ndb_almanac, ndb_almanac_md,
+                      sizeof(almanac_t), PLATFORM_SIGNAL_COUNT)
+            != NDB_ERR_NONE) {
+        /* TODO: log failure */
+    }
+    return NDB_ERR_NONE;
+}
+
+enum ndb_op_code ndb_ephemeris_read(gnss_signal_t *sid, ephemeris_t *e)
```
jacobmcnamee: For better or worse, we're passing SIDs by value everywhere else right now.

src/ndb_p1.c
```diff
@@ -0,0 +1,134 @@
...
+                      &ndb_almanac_md[idx]);
+}
+
+void ndb_p1_nvm_write()
+{
+    int idx;
+    for(idx = 0; idx < PLATFORM_SIGNAL_COUNT; idx++) {
+        if(ndb_ephemeris[idx].valid && !ndb_ephemeris_md[idx].saved){
+            ephemeris_t t_ephe;
+            ndb_lock(1);
```
jacobmcnamee: Can this sequence go into a function comparable to `ndb_update()` or `ndb_retrieve()`?

jacobmcnamee: Or... maybe some way to expose the data elements to the core when you register a section so that it can autonomously scan through them and write to NVM as needed.

src/ndb.c
```diff
@@ -0,0 +1,215 @@
...
+
+enum ndb_op_code ndb_not_implemented()
+{
+    return NDB_ERR_UNSUPPORTED;
+}
+
+enum ndb_op_code ndb_p1_init() OPTF;
```
jacobmcnamee: I think the linked list of structures for each section was good for the `init()` and `nvm_write()` interfaces. You'll always be iterating through all sections when these operations run, anyway.

jacobmcnamee:
Superseded by #630

src/ephemeris.c
```diff
@@ -115,40 +61,24 @@ static void ephemeris_msg_callback(u16 sender_id, u8 len, u8 msg[], void* contex
     return;
   }
 
-  ephemeris_new(&e);
+  u8 v, h;
+  gps_time_t toe;
+  ndb_ephemeris_info(e.sid, &v, &h, &toe, NULL);
+  if (!v || gpsdifftime(&e.toe, &toe) > 0) {
```
jacobmcnamee: If we are moving all the ephemeris candidate stuff into the NDB, then I think this logic should be in the NDB as well.

src/ndb.c
```diff
@@ -0,0 +1,369 @@
...
+#include "sbp.h"
+#include "ndb.h"
+
+#define NDB_THREAD_PRIORITY (LOWPRIO)
+static WORKING_AREA_CCM(ndb_thread_wa, 1756);
+static void ndb_service_thread(void*);
+
+static bool do_writes_to_flash = false;
+
+MUTEX_DECL(data_access);
```
jacobmcnamee: make all these static

src/ndb.c
```diff
@@ -0,0 +1,369 @@
...
+
+enum ndb_op_code ndb_init()
+{
+  if (!do_writes_to_flash) {
+    log_info("NDB: configured not to save data to flash file system");
+  }
+
+  chThdCreateStatic(ndb_thread_wa, sizeof(ndb_thread_wa),
+                    NDB_THREAD_PRIORITY, ndb_service_thread, NULL);
+  ndb_p1_init();
```
jacobmcnamee: Can you add a `platform_ndb_init()` in `platform_signal.c` and put the section init calls there?

src/ndb.c
```diff
@@ -0,0 +1,369 @@
...
+  if (cfs_read(file->fh, ver, sizeof(ndb_file_version))
+      == sizeof(ndb_file_version)) {
+    if (memcmp(&ver, version, sizeof(ndb_file_version)) == 0) {
+      return NDB_ERR_NONE;
+    } else {
+      cfs_close(file->fh);
+      cfs_remove(file->name);
+      return ndb_open_file(file, version);
+    }
+  }
```
jacobmcnamee: Handle case where `cfs_read() != sizeof(ndb_file_version)`

src/ndb.c
```diff
@@ -0,0 +1,369 @@
...
+  return NDB_ERR_NONE;
+}
+
+enum ndb_op_code ndb_wq_get(ndb_element_metadata_t** md)
+{
+  chMtxLock(&wq_access);
+  if (NULL == wq_first) {
+    /* NOTE: chCondwaitTimeout does not care of
+     locking mutex again when timeout occurs*/
+    if (chCondWaitTimeout(&wq_new_data, NV_WRITE_REQ_TIMEOUT) == MSG_TIMEOUT) {
```
jacobmcnamee: I suggest re-checking `wq_first == NULL` instead of relying on the value returned by `chCondWaitTimeout()`

src/ndb.h
```diff
@@ -0,0 +1,145 @@
...
+  almanac_t alma;
+  u32 l2c_capabilities;
+} ndb_element_t;
+
+/* Maximum waiting time for write request, milliseconds */
+#define NV_WRITE_REQ_TIMEOUT 100
+
+#define NDB_IE_DIRTY (1 << 0) /**< needs to be written to NVM */
+#define NDB_IE_VALID (1 << 1) /**< value in RAM has been set */
+#define NDB_IE_IS_VALID(md_ptr) (md_ptr->nv_data.state & NDB_IE_VALID)
```
jacobmcnamee: Recommend parens around `md_ptr`

src/ndb_p1.c
```diff
@@ -0,0 +1,351 @@
...
+  log_info("Should never get here! %s, %d", __FILE__, __LINE__);
+  return NDB_ERR_ALGORITHM_ERROR;
+}
+
+enum ndb_op_code ndb_ephemeris_info(gnss_signal_t sid, u8* v, u8* h,
+                                    gps_time_t* toe, u8* fit_interval)
+{
+  u16 idx = sid_to_global_index(sid);
+  ndb_lock();
+  *v = ndb_ephemeris[idx].valid;
```
jacobmcnamee: I vote making it mandatory to provide valid pointers for all parameters. Otherwise `v` and `h` should also be optional.

src/ndb.c
```diff
@@ -0,0 +1,369 @@
...
+}
+
+void ndb_retrieve(void* out, void* cached, size_t size)
+{
+  ndb_lock();
+  memcpy(out, cached, size);
+  ndb_unlock();
+}
+
+enum ndb_op_code ndb_update(void* cached, void* new, size_t size,
```
jacobmcnamee: What is the purpose of exposing the and/or state masks?

src/ndb.c
```diff
@@ -0,0 +1,369 @@
...
+                            enum ndb_data_source src,
+                            ndb_element_metadata_t *md, u8 and_state_mask,
+                            u8 or_state_mask)
+{
+  ndb_lock();
+  if (memcmp(new, cached, size) == 0) {
+    ndb_unlock();
```
jacobmcnamee: Update timestamp?

src/ndb_p3.c
```diff
@@ -0,0 +1,77 @@
...
+#include "ndb.h"
+
+#define GPS_L2C_CAPB_FILE_NAME "l2capb"
+static u32 gps_l2c_capabilities _CCM;
+static ndb_element_metadata_t gps_l2c_capabilities_md _CCM;
+
+static ndb_file_t gps_l2c_capb_file = {
+  .name = GPS_L2C_CAPB_FILE_NAME,
+  .fh = -1,
+  .expected_size = sizeof(ndb_file_version) +
```
jacobmcnamee: If every file will include version information, I would abstract that away in ndb.c. No reason the individual sections need to know about it.

src/ndb.c
jacobmcnamee: `cached` and `size` are contained in the metadata, right? Do those need to be passed in separately?

src/Makefile
```diff
@@ -112,6 +112,9 @@ CSRC := $(STARTUPSRC) \
         $(SWIFTNAV_ROOT)/src/pps.o \
         $(SWIFTNAV_ROOT)/src/decode.o \
         $(SWIFTNAV_ROOT)/src/signal.o \
+        $(SWIFTNAV_ROOT)/src/ndb.o \
+        $(SWIFTNAV_ROOT)/src/ndb_p1.o \
```
jacobmcnamee: Also make these board-specific. See `board/v[2,3]/Makefile.inc

src/ndb.c
```diff
@@ -0,0 +1,369 @@
...
+    if (data_write_ok) {
+      ndb_lock();
+      changed = memcmp(md->data, &buf, md->data_size);
+      ndb_unlock();
+      if (!changed) {
+        /*
+         * Data hasn't changed while i/o operation was running -
+         * mark as saved and write metadata to NVM.
+         * */
+        md->nv_data.state &= ~NDB_IE_DIRTY;
```
jacobmcnamee: Should metadata modification be inside `ndb_lock()` ?

jacobmcnamee:
Added a bunch of inline comments. A couple other suggestions:

### Separate the internal/public APIs as well as different data types into separate files
E.g.,
- ndb_internal.c/h: internal API
- ndb_almanac.c/h, ndb_ephemeris.c/h, ndb_xxx.c/h: data and public API for specific data types
- ndb.h: include all the individual public API headers (if we really want this?)

src/ndb_p1.c
```diff
@@ -0,0 +1,351 @@
...
+  u32 i;
+
+  memset(ephe_candidates, 0, sizeof(ephe_candidates));
+
+  memset(ndb_ephemeris, 0, sizeof(ndb_ephemeris));
+  for (i = 0; i < PLATFORM_SIGNAL_COUNT; i++) {
+    ndb_ephemeris[i].sid = sid_from_global_index(i);
+  }
+
+  if (ndb_load_data(&ndb_ephe_file, ndb_ephemeris, ndb_ephemeris_md,
```
jacobmcnamee: Add common functions to do file setup, metadata setup, etc.

src/ndb.h
```diff
@@ -0,0 +1,145 @@
...
+  ndb_file_handle_t fh; /**< File handle */
+  u32 expected_size; /**< Expected file size */
+} ndb_file_t;
+
+typedef union
+{
+  ephemeris_t ephe;
+  almanac_t alma;
+  u32 l2c_capabilities;
+} ndb_element_t;
```
jacobmcnamee: I think it would be better to simply allocate a buffer with the required size rather than relying on this union to be updated with every possible element type. E.g. just do `u8 buf[md->data_size]` instead of `ndb_element_t buf`.

rge-exafore: Done

src/Makefile
rge-exafore: I moved src/ndb_p3.o to v2/Makefile.inc. src/ndb_p1.o is common for both boards - is it fine to leave it here?

src/ephemeris.c
rge-exafore: Yes, makes sense as NDB knows the source. I moved that logic to ndb_ephemeris_store().

src/ndb.c
rge-exafore: Done.

src/ndb.c
rge-exafore: Done.

src/ndb.c
rge-exafore: Yes, done.

src/ndb.c
rge-exafore: If file contains less data than sizeof(ndb_file_version) we just reuse that file as it was opened successfully for read/write access. We write there version signature and it's equivalent to creating a new file and doing the same write. Am I wrong somewhere?

src/ndb.h
rge-exafore: Done

src/ndb.c
rge-exafore: Yes, and source too. And trigger saving to flash. Maybe use different flags for dirty data and dirty metadata is metadata can change while data stay the same?

src/ndb.c
rge-exafore: On masks: I though of using those for information element -specific flags. So, that ndb_update gets then from outside from IE-aware function. I cannot use them anywhere at the moment, though. Do you think I should remove them?

src/ndb_p1.c
rge-exafore: Done.

src/ndb_p3.c
rge-exafore: Done

src/ndb.c
rge-exafore: On cached and size - done.

src/ndb_p1.c
rge-exafore: I didn't get that: ndb_load_file is a common function.

jacobmcnamee:
Regarding the missing trailing zeros with CFS: Can you add a footer (magic value, CRC maybe) at the end of all the data / metadata?

src/ndb.c
jacobmcnamee: I guess that's fine. I do think we should probably verify the total file size and maybe add a CRC or something, as well.

src/ndb.c
jacobmcnamee: Flags - I wouldn't combine internal and external flags into the same variable. Maybe just remove the external API for now and we can either add a `user_flags` field or manage them entirely outside of the core ndb.c.

src/ndb.c
jacobmcnamee: Not sure. Though I'm a little worried about excessive writes to flash. I haven't done any calculations on lifetime.

src/ndb_p1.c
jacobmcnamee: I was referring to the whole chunk of code: loading, logging the file status, and setting up the metadata. That's going to be the same for all data types, right?

src/Makefile
jacobmcnamee: What are your thoughts on renaming / moving around some files as I mentioned https://github.com/swift-nav/piksi_firmware/pull/621#issuecomment-211687253 ?

rge-exafore:
I added one non zero byte to the end of file. I don't think adding that to each metadata or data makes sense. We use file of fixed size,, all reads and writes are done at known offset based on sid. So, just one byte at the end of the file should be fine. We can change that later to CRC is needed. Is that approach ok?

src/Makefile
rge-exafore: In progress. I had to distract to cfs fix and rebasing to Leith's changes. Will do changes now according to these requirements.

src/ndb.c
rge-exafore: Done

src/manage.c
```diff
@@ -244,11 +240,14 @@ static u16 manage_warm_start(gnss_signal_t sid, const gps_time_t* t,
...
     float el = 0;
     double el_d, _, dopp_hint = 0, dopp_uncertainty = DOPP_UNCERT_ALMANAC;
+    union { ephemeris_t ephe; almanac_t alma; } orbit;
+    enum ndb_op_code oc;
 
     /* Do we have a suitable ephemeris for this sat?  If so, use
        that in preference to the almanac. */
-    const ephemeris_t *e = ephemeris_get(sid);
-    if (ephemeris_valid(e, t)) {
+    const ephemeris_t *e = &orbit.ephe;
```
jacobmcnamee: Why the `const`?

src/manage.c
```diff
@@ -266,8 +265,9 @@ static u16 manage_warm_start(gnss_signal_t sid, const gps_time_t* t,
       if (time_quality >= TIME_FINE)
         dopp_uncertainty = DOPP_UNCERT_EPHEM;
     } else {
-      const almanac_t *a = &almanac[sid_to_global_index(sid)];
-      if (a->valid) {
+      const almanac_t *a = &orbit.alma;
```
jacobmcnamee: same as ^

src/ndb/ndb_internal.c
```diff
@@ -0,0 +1,426 @@
...
+  return NDB_ERR_NONE;
+}
+
+enum ndb_op_code ndb_wq_get(ndb_element_metadata_t** md)
+{
+  chMtxLock(&wq_access);
+  if (NULL == wq_first) {
+    /* NOTE: chCondwaitTimeout does not care of
+     locking mutex again when timeout occurs*/
+    chCondWaitTimeout(&wq_new_data, NV_WRITE_REQ_TIMEOUT);
```
jacobmcnamee: Double check for which return codes this function leaves the mutex unlocked and make sure there is no way to return in an incorrect state.

src/ndb/ndb_internal.c
```diff
@@ -0,0 +1,426 @@
...
+    u16 n_elements = md->file->n_elements;
+    u8 buf[element_size];
+    ndb_lock();
+
+    assert((md->nv_data.state & (NDB_IE_DIRTY + NDB_MD_DIRTY)) != 0);
+
+    if(md->nv_data.state & NDB_IE_DIRTY)
+      memcpy(&buf, md->data, element_size);
+
+    memcpy(&md_copy, md, sizeof(ndb_element_metadata_t));
```
jacobmcnamee: check NDB_MD_DIRTY flag before copying?

src/ndb/ndb_internal.c
```diff
@@ -0,0 +1,426 @@
...
+
+  if (NDB_ERR_TIMEOUT == ndb_wq_get(&md)) {
+    return NDB_ERR_TIMEOUT;
+  } else {
+    if (!do_writes_to_flash) {
+      *data_write_ok = *md_write_ok = true;
+      return NDB_ERR_NONE;
+    }
+    ndb_element_metadata_t md_copy;
+    u16 element_size = md->file->data_size;
```
jacobmcnamee: Are these fields required to be constant? Otherwise should they be copied while locked?

src/ndb/ndb_internal.c
```diff
@@ -0,0 +1,426 @@
...
+#include "ndb_internal.h"
+
+#define NDB_THREAD_PRIORITY (LOWPRIO)
+static WORKING_AREA_CCM(ndb_thread_wa, 1756);
+static void ndb_service_thread(void*);
+
+u8 ndb_file_version[MAX_NDB_FILE_VERSION_LEN];
+u8 ndb_file_end_mark = 0xb6;
+
+static bool do_writes_to_flash = false;
```
jacobmcnamee: Should this be a compile-time define? It seems like things will break if it is changed from false to true at runtime.

src/board/v2/platform_signal.c
```diff
@@ -25,3 +27,10 @@ void platform_decode_setup(void)
 {
   decode_gps_l1ca_register();
 }
+
+void platform_ndb_init(void)
+{
+  ndb_init();
```
jacobmcnamee: Track and decode do:
- <module>_init() from main.c
- platform_<module>_init() from <module>_init()

src/board/v3/platform_signal.c
```diff
@@ -25,3 +27,11 @@ void platform_decode_setup(void)
 {
   decode_gps_l1ca_register();
 }
+
+void platform_ndb_init(void)
+{
+  ndb_init();
+  ndb_p1_init();
```
jacobmcnamee: Just call the individual ndb_\<type\>_init() functions directly here. Keep the board-specific abstraction in this file.

jacobmcnamee: And I guess do the same for the sbp_update() stuff.

src/ndb/ndb_almanac.c
```diff
@@ -0,0 +1,81 @@
...
+  if (md->update_c == *uc) {
+    return NDB_ERR_NONE;
+  }
+
+  enum ndb_op_code r = ndb_almanac_read(cached_a->sid, cached_a);
+  if (NDB_ERR_NONE != r) {
+    return r;
+  }
+
+  *uc = md->update_c;
```
jacobmcnamee: I think the update count check / set and data read all need to be atomic.

src/ndb/ndb_almanac.c
```diff
@@ -0,0 +1,81 @@
...
+  return NDB_ERR_NONE;
+}
+
+enum ndb_op_code ndb_almanac_store(almanac_t *a, enum ndb_data_source src)
+{
+  u16 idx = sid_to_global_index(a->sid);
+  return ndb_update(a, src, &ndb_almanac_md[idx]);
+}
+
+enum ndb_op_code ndb_update_cache_almanac(almanac_t *cached_a,
```
jacobmcnamee: Rename to ndb_almanac_cache_update() for consistency?

src/ndb/ndb_internal.c
```diff
@@ -0,0 +1,426 @@
...
+static enum ndb_op_code ndb_wq_get(ndb_element_metadata_t** md);
+
+static enum ndb_op_code ndb_open_file(ndb_file_t *file);
+static enum ndb_op_code ndb_read(ndb_file_t *f, cfs_offset_t o,
+                                 void *first_element, size_t s);
+
+static enum ndb_op_code do_nv_writes(bool *data_write_ok, bool *md_write_ok);
+static void do_sbp_updates();
+
+enum ndb_op_code ndb_gps_l2cm_l2c_cap_store(u32 *l2c_cap,
```
jacobmcnamee: These can be removed.

jacobmcnamee: These can be removed from this file.

src/ndb/ndb_internal.c
```diff
@@ -0,0 +1,426 @@
...
+  memset(data, 0, ds);
+  memset(metadata, 0, sizeof(ndb_element_metadata_t) * el_number);
+
+  if (do_writes_to_flash && (f->fh >= 0)) {
+    memset(md_nv, 0, mds);
+    if (ndb_write_file_data(f, 0, data, ds) == NDB_ERR_NONE) {
+      if (ndb_write_file_data(f, ds, md_nv, mds) == NDB_ERR_NONE) {
+        if (ndb_write_file_data(f, ds + mds,
+                                &ndb_file_end_mark, 1) == NDB_ERR_NONE) {
+          r = NDB_ERR_INIT_DONE;
```
jacobmcnamee: `r` will be set to NDB_ERR_NONE if the top block succeeds but the bottom block does not.

src/manage.c
rge-exafore: removed.

src/manage.c
rge-exafore: removed.

src/ndb/ndb_almanac.c
rge-exafore: Renamed both almanac- and ephemeris-related functions.

src/ndb/ndb_almanac.c
rge-exafore: Updated both almanac- and ephemeris-related functions.

src/ndb/ndb_internal.c
rge-exafore: Done.

src/ndb/ndb_internal.c
rge-exafore: That looks weird to me, but the documentation says:
 * @post    Exiting the function because a timeout does not re-acquire the mutex, the mutex ownership is lost.

src/ndb/ndb_internal.c
rge-exafore: Fixed.

src/ndb/ndb_internal.c
rge-exafore: I had that initially, but then thought that as metadata are to be copied and saved always, that check would be redundant. Even NDB_MD_DIRTY flag can be dropped. I just want to keep it so that there is clear indication what caused the write request.

src/ndb/ndb_internal.c
rge-exafore: Yes, they are const. I changed ndb_file_t declaration.

src/ndb/ndb_internal.c
rge-exafore: Right. Sorry, forgot to discuss with that. This is temporary implementation. In theory we can implement a function that opens a file, saves data from ram and changes this flag to true. However, I don't see any need for that. So, I see two options: compile time define (the decision is not to use those, right?) or weak functions for file access returning success of open/close/seek/write and zeros on read. Which option is to be taken?

src/board/v2/platform_signal.c
rge-exafore: Added ndb.c with ndb_setup(). Now it fully matches the pattern, IMO.

src/board/v3/platform_signal.c
rge-exafore: Done.

src/ndb/ndb_ephemeris.c
```diff
@@ -0,0 +1,273 @@
...
+} ephemeris_candidate_t;
+
+typedef enum {
+  EPHE_IDENTICAL,
+  EPHE_NEW_CANDIDATE,
+  EPHE_NEW_TRUSTED,
+  EPHE_CAND_MISMATCH,
+} ndb_ephemeris_status_t;
+
+#define EPHE_CAND_LIST_LEN (MAX_CHANNELS*2/3)
```
jacobmcnamee: Why the 2/3?

src/ndb/ndb_ephemeris.c
```diff
@@ -0,0 +1,273 @@
...
+      {
+        u16 idx = sid_to_global_index(e->sid);
+        return ndb_update(e, src, &ndb_ephemeris_md[idx]);
+      }
+      case EPHE_NEW_CANDIDATE:
+      case EPHE_CAND_MISMATCH:
+        return NDB_ERR_UNRELIABLE_DATA;
+    }
+  } else if (NDB_DS_SBP == src) {
+#ifndef NEVER_DEFINED
```
jacobmcnamee: Can this be removed?

src/ndb/ndb_internal.c
jacobmcnamee: In that case I guess you should return an error, right? Otherwise if the mutex is reacquired check the write queue as you are now.

jacobmcnamee: Just make sure that 
1. The mutex is always handled properly.
2. You continue to check the write queue explicitly and retry or return an error if necessary, regardless of what the cond var does.

src/signal.c
```diff
@@ -120,7 +120,7 @@ gnss_signal_t sid_from_constellation_index(enum constellation constellation,
  */
 u16 sid_to_global_index(gnss_signal_t sid)
 {
-  assert(code_supported(sid.code));
```
jacobmcnamee: I think this change is incorrect. `sid_to_code_index()` asserts that the SID is valid, and the global index is only defined for supported codes.

src/ndb/ndb_internal.c
jacobmcnamee: If it's reasonable to split the NVM read / write functions into a separate file, maybe have one file that accesses CFS and another that does nothing. Then include the desired implementation in Makefile.inc. Otherwise just use a #define somewhere. I'd rather do either of those than have a runtime option (variable) which can break things if changed dynamically.

jacobmcnamee: I anticipate some refactoring here when we add NVM support to v3, so I'm not super concerned right now.

src/signal.c
rge-exafore: Yes, you are right. I somehow missed sid_to_code_index() call. Reverted.

src/ndb/ndb_ephemeris.c
rge-exafore: Yes, it will be removed eventually. It's crash debugging part.

src/ndb/ndb_ephemeris.c
rge-exafore: Just to save memory. Actually, with latest improvement memory isn't a concern, so I change it to MAX_CHANNELS.

src/ndb/ndb_internal.c
rge-exafore: Ok, I went ahead and added ndb_fs_access.c to 'v2' and 'v3' directories. At the moment they don't implement real access to file system, but that can easily be changed. Please review this part.

src/ndb/ndb_internal.c
rge-exafore: I changes the implementation to always check the pointer. Also added couple of asserts to see if anything goes wrong.

src/board/v3/ndb_fs_access.c
```diff
@@ -0,0 +1,102 @@
...
+#ifdef USE_CFS
+  return cfs_write(fd, buf, size);
+#else
+  (void)fd;
+  (void)buf;
+  return size;
+#endif
+}
+
+int ndb_coffee_reserve(const char *name, cfs_offset_t size)
```
jacobmcnamee: `ndb_fs_reserve()` ?

jacobmcnamee:
FS split looks good. Whenever you're done testing just remove the `USE_CFS` #define and only include the desired code path for each board's implementation.

rge-exafore:
I removed USE_CFS from v2 version. However, I propose to keep it in v3 until nvm filesystem works there.

src/ndb/ndb_internal.h
```diff
@@ -0,0 +1,83 @@
...
+                   ndb_element_metadata_t* metadata,
+                   const size_t el_size,
+                   const size_t el_number);
+enum ndb_op_code ndb_update(void* new, enum ndb_data_source src,
+                            ndb_element_metadata_t *md);
+void ndb_retrieve(void* out, void* cached, size_t size);
+enum ndb_op_code ndb_write_file_data(ndb_file_t *f, cfs_offset_t o, void *b,
+                                     size_t l);
+
+void ndb_p1_sbp_update();
```
jacobmcnamee: These can be removed

src/ndb/ndb_weaks.c
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+#include "ndb.h"
+#include "ndb_internal.h"
+
+
+enum ndb_op_code ndb_gps_l2cm_l2c_cap_store(u32 *l2c_cap,
```
jacobmcnamee: Just add `NDB_WEAK` to the prototypes in the header files and you shouldn't need this file at all.

jacobmcnamee: I think all the read/store/update functions should have `NDB_WEAK` and the init/sbp function should not.

src/ndb/ndb_internal.h
rge-exafore: Done.

src/ndb/ndb_common.h
```diff
@@ -0,0 +1,55 @@
...
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef SRC_NDB_COMMON_H_
+#define SRC_NDB_COMMON_H_
+
+#include <libswiftnav/common.h>
+#include <libswiftnav/ephemeris.h>
```
jacobmcnamee: What do you think about moving the datatype-specific includes to the respective header files?

jacobmcnamee:
Does v3 crash with USE_CFS defined? It should function normally aside from the volatile nature of running from RAM.

PR #618, opened 2016-02-19
WIP: Port tracking loops for L2C

dt-exafore:
This ports tracking loop for L2CM from Peregrine

PR #622, opened 2016-02-29
WIP: L2C handover porting

dt-exafore:
Just a Plan for porting L2C handover with TODOs that will be implemented after clarification.


dt-exafore:
This will be implemented under https://github.com/swift-nav/piksi_firmware/pull/618

PR #624, opened 2016-03-04
Tracking channel refactor, specialize by code type

jacobmcnamee:
Major overhaul of tracking module:
- No more global `tracking_channel[]`
- Code-specific tracking init and update implementations
- API for code-specific implementations to reuse common data / functions
- Thread safety improvements (copying out measurements, handling race condition w/NAP interrupts)
- Tracking startup queue to enable hot starting new channels from within the tracking loops

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -90,6 +90,29 @@ static bool track_mask[PLATFORM_SIGNAL_COUNT];
 #define DOPP_UNCERT_ALMANAC 4000
 #define DOPP_UNCERT_EPHEM   500
 
+#define COMPILER_BARRIER() asm volatile ("" : : : "memory")
+
+#define TRACKING_STARTUP_FIFO_SIZE 8
+
+#define TRACKING_STARTUP_FIFO_INDEX_MASK ((TRACKING_STARTUP_FIFO_SIZE) - 1)
+#define TRACKING_STARTUP_FIFO_INDEX_DIFF(write_index, read_index) \
+          ((tracking_startup_fifo_index_t)((write_index) - (read_index)))
+#define TRACKING_STARTUP_FIFO_LENGTH(p_fifo) \
+          (TRACKING_STARTUP_FIFO_INDEX_DIFF((p_fifo)->write_index, \
+                                            (p_fifo)->read_index))
+
+typedef u8 tracking_startup_fifo_index_t;
+
+typedef struct {
+  tracking_startup_fifo_index_t read_index;
+  tracking_startup_fifo_index_t write_index;
+  tracking_startup_params_t elements[TRACKING_STARTUP_FIFO_SIZE];
+} tracking_startup_fifo_t;
+
+static tracking_startup_fifo_t tracking_startup_fifo;
+
+static MUTEX_DECL(tracking_startup_mutex);
+
 static almanac_t almanac[PLATFORM_SIGNAL_COUNT];
 
 static float elevation_mask = 0.0; /* degrees */
@@ -99,6 +122,14 @@ static u8 manage_track_new_acq(gnss_signal_t sid);
 static void manage_acq(void);
 static void manage_track(void);
 static void nmea_send(void);
+
+static void manage_tracking_startup(void);
+static void tracking_startup_fifo_init(tracking_startup_fifo_t *fifo);
+static bool tracking_startup_fifo_write(tracking_startup_fifo_t *fifo,
+                                        const tracking_startup_params_t *
+                                        element);
+static bool tracking_startup_fifo_read(tracking_startup_fifo_t *fifo,
+                                       tracking_startup_params_t *element);
 
 static sbp_msg_callbacks_node_t almanac_callback_node;
 static void almanac_callback(u16 sender_id, u8 len, u8 msg[], void* context)
@@ -141,6 +172,7 @@ static msg_t manage_acq_thread(void *arg)
   chRegSetThreadName("manage acq");
   while (TRUE) {
     manage_acq();
+    manage_tracking_startup();
     watchdog_notify(WD_NOTIFY_ACQ_MGMT);
   }
 
@@ -150,6 +182,8 @@ static msg_t manage_acq_thread(void *arg)
 void manage_acq_setup()
 {
   SETTING("acquisition", "sbas enabled", sbas_enabled, TYPE_BOOL);
+
+  tracking_startup_fifo_init(&tracking_startup_fifo);
 
   for (u32 i=0; i<PLATFORM_SIGNAL_COUNT; i++) {
     acq_status[i].state = ACQ_PRN_ACQUIRING;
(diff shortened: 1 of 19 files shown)


src/manage.h
```diff
@@ -58,6 +62,14 @@
 #define MANAGE_TRACK_THREAD_PRIORITY (NORMALPRIO-2)
 #define MANAGE_TRACK_THREAD_STACK   1400
 
+typedef struct {
+  u32 sample_count;
+  float carrier_freq;
+  float carrier_phase;
```
rge-exafore: should be code phase

src/manage.c
```diff
@@ -381,28 +389,9 @@ static void manage_acq()
...
-  /* Start the tracking channel */
-  tracking_channel_init(chan, acq->sid, cf, track_count, cn0,
-                        TRACKING_ELEVATION_UNKNOWN);
-  /* TODO: Initialize elevation from ephemeris if we know it precisely */
-
-  /* Start the decoder channel */
-  if (!decoder_channel_init(chan, acq->sid)) {
-    log_error("decoder channel init failed");
+  } else {
+    acq->state = ACQ_PRN_TRACKING;
```
rge-exafore: Isn't it better to not remember on acquisition side which SVs are being tracked and just query that from tracking in choose_acq_sat function (that query won't require heavy thread synchronization as it's about atomic reading of status field)? Then here acquisition can just blindly form a tracking_startup_params_t structure and throw it to tracking. If we put tracking_startup_params_t structure as a field to acq_status_t and add a link to the next tracking_startup_params_t we can use the same technique as in other places to form a queue of requests to tracking. This would improve inter-thread communication IMO as won't be stopping threads for performing potentially long operations (like tracking start).

src/manage.c
```diff
@@ -90,13 +90,15 @@ static acq_status_t acq_status[PLATFORM_SIGNAL_COUNT];
 
 static almanac_t almanac[PLATFORM_SIGNAL_COUNT];
```
rge-exafore: That will go as we take Nav DB into use.

src/manage.c
```diff
@@ -462,9 +451,6 @@ static msg_t manage_track_thread(void *arg)
 
 void manage_track_setup()
```
rge-exafore: Splitting to manage_acq.c and manage_track.c would make the code more readable IMO

src/manage.c
```diff
@@ -560,30 +541,42 @@ static void manage_track()
...
+  for (u32 i=0; i<nap_track_n_channels; i++) {
+    if (tracking_channel_running(i)) {
+      gnss_signal_t sid = tracking_channel_sid_get(i);
+      if (sid_to_constellation(sid) == CONSTELLATION_GPS) {
+        prns[num_prns++] = sid.sat;
+      }
+    }
+  }
+  /* Send GPGSA message */
+  nmea_gpgsa(prns, num_prns, 0);
```
rge-exafore: Sending NMEA would be better to put to a separate task perhaps?

src/manage.c
jacobmcnamee: As soon as we the almanac is in the NDB I'm all for this.

src/manage.h
jacobmcnamee: fixed

jacobmcnamee:
/cc @swift-nav/firmware 
Review at your peril. But please review.

src/manage.c
```diff
@@ -610,4 +635,145 @@ u8 tracking_channels_ready()
...
+            SCORE_ACQ + (startup_params.cn0_init - ACQ_THRESHOLD);
+        acq->dopp_hint_low = startup_params.carrier_freq - ACQ_FULL_CF_STEP;
+        acq->dopp_hint_high = startup_params.carrier_freq + ACQ_FULL_CF_STEP;
+      }
+
+      continue;
+    }
+
+    /* Transition to tracking. */
+    u32 track_count = nap_timing_count() + 20000;
```
dt-exafore: nap_timing_count() returns u64, could you please clarify do we need to store 32 LSB here only?

src/track.h
```diff
@@ -18,113 +19,63 @@
...
- * We have to declare the number here as the number of tracking channels in
- * the FPGA is read at runtime. */
-/* TODO: NAP_MAX_N_TRACK_CHANNELS is a duplicate of MAX_CHANNELS */
-extern tracking_channel_t tracking_channel[NAP_MAX_N_TRACK_CHANNELS];
+void track_setup(void);
 
-void initialize_lock_counters(void);
+void tracking_send_state(void);
 
 float propagate_code_phase(float code_phase, float carrier_freq, u32 n_samples);
```
dt-exafore: this function uses following, which is GPS_L1 specific:
```
  /* Calculate the code phase rate with carrier aiding. */
  u32 code_phase_rate = (1.0 + carrier_freq/GPS_L1_HZ) *
                            NAP_TRACK_NOMINAL_CODE_PHASE_RATE;
```
I remember we agreed I will replace the magic numbers like GPS_L1_HZ with a function that takes sid as parameter. The question is how to do this in this function? There is no any SID passed to it.

jacobmcnamee: This is old code which is probably only showing up as changed since it was moved around. I imagine we need a) a function in libswiftnav which returns nominal carrier frequency by code type, and/or b) to store the nominal carrier frequency in the `tracking_startup_params_t` structure. I'm leaning toward using the lookup function only for acquisition and then storing that value in the startup params / tracking channel structs for subsequent use.

src/manage.c
jacobmcnamee: Old code I just relocated. But it looks like the upper 32 bits are only maintained in software
```
/** Get the current NAP internal sample clock count.
 * NAP's internal count of sample clocks + (number of NAP's counter rollovers)
 * times 2^32. NAP's internal sample clock counter is 32 bits wide - at a
 * 16.368MHz sample clock frequency it rolls over approximately every 262
 * seconds.
 *
 * \return NAP's internal count of sample clocks +
 *               (total number of NAP counter rollovers) * 2^32.
 */
u64 nap_timing_count(void)
{
  static u32 rollover_count = 0;
  static u32 prev_count = 0;

  u8 temp[4] = { 0, 0, 0, 0 };

  nap_xfer_blocking(NAP_REG_TIMING_COUNT, 4, temp, temp);

  u32 count = (temp[0] << 24) | (temp[1] << 16) | (temp[2] << 8) | temp[3];

  if (count < prev_count)
    rollover_count++;

  prev_count = count;

  return (u64)count | ((u64)rollover_count << 32);
}
```

src/manage.c
```diff
@@ -88,15 +90,46 @@ static acq_status_t acq_status[PLATFORM_SIGNAL_COUNT];
 #define DOPP_UNCERT_ALMANAC 4000
 #define DOPP_UNCERT_EPHEM   500
 
+#define COMPILER_BARRIER() asm volatile ("" : : : "memory")
+
+#define TRACKING_STARTUP_FIFO_SIZE 8
```
rge-exafore: @jacobmcnamee: I'd propose to add a comment that TRACKING_STARTUP_FIFO_SIZE must be power of 2 as otherwise the algorithm won't work correctly.

jacobmcnamee: done

PR #625, opened 2016-03-08
solution: Remove dependence on hardware timer.

gsmcmullin:
Use ChibiOS system tick to schedule solution instead.

/cc @swift-nav/firmware


ljbade:
Hmm must have missed this earlier. 

Looks good for merge

PR #629, opened 2016-03-18
Add file name and line number to log output

dt-exafore:
The changes done in order to support new log output.
NOTE: PR #314 (libswiftnav) must be merged first.


dt-exafore:
@rge-exafore, please, review.

PR #632, opened 2016-03-29
Dual target STM32 and Zynq, ChibiOS update

jacobmcnamee:
- Upgrade to ChibiOS/RT 3
- Use ChibiOS/HAL drivers
- Remove libopencm3
- Zynq (v3) build target
- SBP TX optimizations
- Tracking update optimizations


jacobmcnamee:
Requires swift-nav/libswiftnav#319

PR #634, opened 2016-03-31
Fix sid_from_sbp() for GPS L1CA 1

jacobmcnamee:
`sid_valid()` would return false for `sid.sat = 0` and hence the offset would incorrectly not be applied.


ljbade:
lgtm

PR #636, opened 2016-04-02
NMEA Fixes

denniszollo:
NMEA Fixes:
Add in ZDA message
Fix COG as 0 - 360
Populate Age of corrections as station ID

TODO:  Document station ID (lower bits)
             Document GEOD offset
             Add in setting for the rate of the tracking status message equivalent in NMEA
             fix RTK DOP https://github.com/swift-nav/piksi_firmware/blob/master/src/solution.c#L185

### DIFF
--- a/src/nmea.c
+++ b/src/nmea.c
@@ -18,8 +18,10 @@
 #include <libswiftnav/coord_system.h>
 #include <libswiftnav/constants.h>
 #include <libswiftnav/logging.h>
+#include <libswiftnav/time.h>
 
 #include "board/nap/track_channel.h"
+#include "track.h"
 #include "nmea.h"
 #include "peripherals/usart.h"
 #include "sbp.h"
@@ -32,6 +34,7 @@ u32 gprmc_msg_rate = 10;
 u32 gpvtg_msg_rate =  1;
 u32 gpgll_msg_rate = 10;
 u32 gpzda_msg_rate = 10;  //!!
+u32 gpgsa_msg_rate = 10;
 static struct nmea_dispatcher *nmea_dispatchers_head;
 /** \addtogroup io
  * \{ */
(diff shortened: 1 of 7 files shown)


denniszollo:
I'm having trouble deciding how to populate a real DOP for the RTK pseudo-absolute position here: https://github.com/swift-nav/piksi_firmware/blob/master/src/solution.c#L185 both mathematically and form a software architecture perspective.

- Possible options: call calc PVT again on the midpoint with a subset of sats and no raim
- send the SPP DOP and document that it is wrong
- bring out the DOPS as part of the baseline solving stuff (might still involve another calc_pvt)

I was hoping that our NMEA outputs would be "complete" with the next release (hence this PR).  

Anyone have some advice?
@fnoble @scottswift @benjamin0 @jacobmcnamee

ljbade:
Shouldn't the DOP be computed from the final set of satellites used in the RTK baseline?

denniszollo:
@ljbade that would be the correct way to compute the dop, but it would require hella firmware and libswiftnav refactoring.  Instead I'm going to use the SPP DOPS but clearly put that the DOPS isn't strictly correct in our documentation and release notes.    It's still better than a hard-coded value.

src/nmea.c
```diff
@@ -29,8 +30,10 @@ const char NMEA_MODULE[] = "nmea";
 
 u32 gpgsv_msg_rate = 10;
 u32 gprmc_msg_rate = 10;
-u32 gpvtg_msg_rate = 10;
```
denniszollo: Course over ground at full message rate now

denniszollo:
This is ready for code review @jacobmcnamee @gsmcmullin : 

I am a little concerned about addressing the tracking global state outside of the tracking thread. @jacob, can I do this?

I ended up using the SPP DOP and will note that in the documentation. 

It requires documentation update in our new NMEA wiki page, as well as settings.yaml for the console.

ljbade:
Hmm interesting, we should have a dop fiction that just takes a receiver position and a set of ephemeris

denniszollo:
TODO: get the RMC message to reflect current fix mode

jacobmcnamee:
lgtm

denniszollo:
@jacobmcnamee 
I added in a lock of tracking channels to avoid sending out false data - let me know if you see any issues with this.  I ran some HITL tests of this and the lock didn't seem to cause any errors.

jacobmcnamee:
lgtm, though eventually I would like to refactor the tracking channel interface such that locks are only used internally. E.g. there could be a function which atomically populates an array with all actively tracked SIDs.

PR #639, opened 2016-04-07
Gracefully handle NAP errors

jacobmcnamee:
Set a flag if a NAP update is missed so that the channel may be dropped.

Also improves handling of NAP interrupts after disabling a tracking channel.


jacobmcnamee:
\cc @swift-nav/firmware

PR #642, opened 2016-04-08
Add timeout to spp when pseudo-absolute output enabled

denniszollo:
/cc @jacobmcnamee @fnoble 

This has been causing a headache for users for ever that we continue to output SPP when a psuedo-absolute solution is available.  In this PR I copy the global state that existed for NMEA outputs to have a timeout to only output SPP when the RTK pseudo abs solution has been unavailable for a certain amount of time.  Would like to get this in for the upcoming release.


ljbade:
Interesting, I guess this shows a single KF that somehow combines SPP, float and fix solutions into a single "solution" would be very useful.

jacobmcnamee:
Code looks fine. I can't comment on the implications for users, though.

PR #643, opened 2016-04-08
updated radio strong- no mavlink

rai-swiftnav:
Changed the default AT command in the firmware. New command turns off communication from Mavlink to Raw Data.


denniszollo:
Reword the commit message for "radio string"

PR #646, opened 2016-04-13
Change ram4 memory segment length for Piksi v2

rge-exafore:
Part of CCM memory is allocated for mstack and
  that has to be taken into account in declaration of
  ram4 memory segment where other data are placed.


rge-exafore:
@swift-nav/firmware : please review

jacobmcnamee:
I was going to ask why we can't just alias the main stack to ram4, but then I remembered that the bootloader (somewhat undesirably) checks the initial stack pointer in the vector table against an expected value:
```
#define STACK_ADDRESS 0x10010000
```
So I guess this is the best way to deal with that.

rge-exafore:
@jacobmcnamee, is that lgtm and ok to merge?

PR #647, opened 2016-04-13
Enable and use backup RAM

jacobmcnamee:
- Changes around clock / HAL init to enable backup RAM
- Move acq and track thread stacks to backup RAM
- Reduce decode thread size

Should free > 4kB of main RAM


gsmcmullin:
Looks good.  Go ahead and merge as long as you've tested.

PR #648, opened 2016-04-19
Tracking on Piksi v3

gsmcmullin:


src/board/v2/nap/track_channel.c
```diff
@@ -14,44 +14,52 @@
...
+#include "track.h"
 
 #include "libswiftnav/prns.h"
+#include "libswiftnav/constants.h"
 
-/** \addtogroup nap
- * \{ */
-
-/** \defgroup track_channel Track Channel
- * Interface to the SwiftNAP track channels.
```
jacobmcnamee: Add this back? Otherwise there are two section ends at the end of the file.

src/board/v2/nap/track_channel.c
```diff
@@ -62,34 +70,38 @@ void nap_track_init_pack(u8 pack[], u8 prn, s32 carrier_phase, u16 code_phase)
...
+
+  double cp_rate = (1 + carrier_freq/GPS_L1_HZ) * GPS_CA_CHIPPING_RATE;
+  nap_track_update(channel, carrier_freq, cp_rate, 0, 0);
+
+  /* Schedule the timing strobe for start_sample_count. */
+  track_count -= SAMPLE_FREQ / (2*GPS_CA_CHIPPING_RATE);
+  nap_timing_strobe(track_count);
+  nap_timing_strobe_wait(100);
+  
+  s->count_snapshot = track_count;
```
jacobmcnamee: Need to ensure that these are stored before before the first interrupt occurs and `nap_track_update()` is called. Probably want
1. Writes to `s`
2. Compiler barrier
3. `nap_timing_strobe()`

jacobmcnamee: I think this might have lead to some of the RAIM issues on hitl.

src/board/v3/nap/nap_common.c
```diff
@@ -10,19 +10,112 @@
...
 
+#define PROCESS_PERIOD_ms (1000)
+
+void nap_isr(void *context);
+
+static BSEMAPHORE_DECL(nap_exti_sem, TRUE);
+static WORKING_AREA_CCM(wa_nap_exti, 2000);
+static void nap_exti_thread(void *arg);
+
+u8 nap_track_n_channels = 1;
```
jacobmcnamee: Initialize to zero?

src/board/v3/nap/nap_common.c
```diff
@@ -10,19 +10,112 @@
...
+
+static void handle_nap_exti(void)
+{
+  u32 irq = NAP->TRK_IRQ;
+
+  while (irq) {
+    tracking_channels_update(irq);
+    NAP->TRK_IRQ = irq;
+
+    asm("dsb");
```
jacobmcnamee: What is this for?

src/board/v3/nap/nap_common.c
```diff
@@ -10,19 +10,112 @@
...
+
+  /* Wake up processing thread */
+  chBSemSignalI(&nap_exti_sem);
+
+  chSysUnlockFromISR();
 }
+
+static void handle_nap_exti(void)
+{
+  u32 irq = NAP->TRK_IRQ;
```
jacobmcnamee: Still a little worried about the slim chance that the TRK_IRQ register reads zero but the IRQ line does not go low fast enough before being asserted again. I guess that's an HDL issue.

src/board/v3/nap/nap_hw.h
```diff
@@ -15,21 +15,48 @@
 
 #include <stdint.h>
 
+/** Max number of tracking channels NAP configuration will be built with. */
+#define NAP_MAX_N_TRACK_CHANNELS     32
+
 typedef struct {
+  volatile uint32_t STATUS;
+  volatile uint32_t CONTROL;
+
```
jacobmcnamee: Is there supposed to be a pattern for the spacing / grouping?

src/board/v3/nap/nap_hw.h
```diff
@@ -68,6 +98,11 @@ typedef struct {
 #define NAP_ACQ_FFT_CONFIG_SCALE_Pos (1U)
 #define NAP_ACQ_FFT_CONFIG_SCALE_Msk (0x3FFFFFFFU << NAP_ACQ_FFT_CONFIG_SCALE_Pos)
 
+#define NAP_TRK_CONTROL_PRN_Pos (3u)
```
jacobmcnamee: Is it PRN or more generically SAT?

src/board/v3/nap/track_channel.c
```diff
@@ -0,0 +1,152 @@
...
+
+  u32 cp_rate = (1.0 + carrier_freq/GPS_L1_HZ) * GPS_CA_CHIPPING_RATE *
+                NAP_TRACK_CODE_PHASE_RATE_UNITS_PER_HZ;
+
+  NAP->TRK_CH[channel].CARR_PINC = -carrier_freq *
+                                   NAP_TRACK_CARRIER_FREQ_UNITS_PER_HZ;
+  NAP->TRK_CH[channel].CODE_PINC = cp_rate;
+  nap_ch_state[channel].len = NAP->TRK_CH[channel].LENGTH = calc_length_samples(1, 0, cp_rate)+1;
+  nap_ch_state[channel].code_phase = (NAP->TRK_CH[channel].LENGTH) * cp_rate;
+  NAP->TRK_CONTROL |= (1 << channel); /* Set to start on the timing strobe */
```
jacobmcnamee: Add a compiler barrier before enabling the channel to make sure that all the `nap_ch_state` writes are done prior.

src/board/v3/nap/track_channel.c
```diff
@@ -0,0 +1,152 @@
...
+}
+
+void nap_track_read_results(u8 channel,
+                            u32* count_snapshot, corr_t corrs[],
+                            double *code_phase_early,
+                            double *carrier_phase)
+{
+  corr_t lc[5];
+  if (NAP->TRK_CH[channel].STATUS & 0x3F)
+    log_warn("Track correlator overflow 0x%08X on channel %d",
```
jacobmcnamee: Can we remove and/or deal with this?

src/track.c
```diff
@@ -349,9 +334,8 @@ bool tracker_channel_init(tracker_channel_id_t id, gnss_signal_t sid,
      * start the channel on an EARLY code phase rollover.
```
jacobmcnamee: Remove comments

src/track/track_gps_l1ca.c
```diff
@@ -70,11 +70,6 @@ static bool use_alias_detection = true;
 
 typedef struct {
   aided_tl_state_t tl_state;   /**< Tracking loop filter state. */
-  u32 code_phase_rate_fp;      /**< Code phase rate in NAP register units. */
```
jacobmcnamee: This is beautiful :)

src/track/track_gps_l1ca.c
```diff
@@ -204,30 +195,21 @@ static void tracker_gps_l1ca_update(const tracker_channel_info_t *channel_info,
...
   } else {
     tracker_correlations_read(channel_info->context, data->cs,
-                              &data->corr_sample_count);
+                              &common_data->sample_count,
+                              &common_data->code_phase_early,
+                              &common_data->carrier_phase);
     alias_detect_first(&data->alias_detect, data->cs[1].I, data->cs[1].Q);
   }
-
-  common_data->sample_count += data->corr_sample_count;
```
jacobmcnamee: :D

src/track_api.h
```diff
@@ -41,9 +41,9 @@ typedef struct {
                                     phase detector last "unlocked". */
   s32 TOW_ms;                  /**< TOW in ms. */
   u32 sample_count;            /**< Total num samples channel has tracked for. */
-  u32 code_phase_early;        /**< Early code phase. */
+  double code_phase_early;        /**< Early code phase. */
```
jacobmcnamee: tabs

src/board/v3/nap/nap_hw.h
wltr: Should probably be SAT. I'm also going to change the naming in the register map to match this.

src/board/v3/nap/nap_common.c
gsmcmullin: Data synchronisation buffer.  This ensures that our write to TRK_IRQ has actually been committed before we try to read it back, in-case the CPU is trying to do something smart.  I added this earlier when we were having trouble with interrupts not arriving.  It may not be required.

src/board/v3/nap/track_channel.c
gsmcmullin: We can remove it, but the scaling should be changed in the HDL so that it doesn't overflow.  I still think we'd like to know if it does.

gsmcmullin:
I'm ready to merge this when you are @jacobmcnamee

src/board/v3/nap/nap_common.c
jacobmcnamee: I don't think it is needed, except possibly as a hack to provide additional delay between the write and the subsequent read (which I would call an FPGA issue if it exists). It's a pointer to volatile so the CPU / compiler shouldn't be doing anything interesting. Also the entire FPGA address space is non-cacheable and strongly ordered, not that even those should matter.

PR #650, opened 2016-04-21
Timeout related to solution rate

denniszollo:
After putting in the fixed timeout for PR #642 , I realized that the timeout needed to be parameterized by the solution rate.  This PR says that the SBP SPP output will be suppressed until 2 solution intervals have passed without a psuedo-absolute position.  It should not change any behavior when not operating in psuedo-absolute mode.

@jacobmcnamee 
Please review. I'm not sure if my use of the preprocessor is advised from a style perspective.

### DIFF
--- a/src/solution.c
+++ b/src/solution.c
@@ -42,6 +42,13 @@
 #include "signal.h"
 #include "system_monitor.h"
 
+/** number of solution periods before SPP resumes in pseudo-absolute mode */
+#define DGNSS_TIMEOUT_PERIODS 2
+
+/** number of OS ticks before SPP resumes in pseudo-absolute mode */
+#define DGNSS_TIMEOUT(soln_freq_hz) MS2ST((DGNSS_TIMEOUT_PERIODS * \
+  1/((float) (soln_freq_hz)) * 1000))
+
 MemoryPool obs_buff_pool;
 mailbox_t obs_mailbox;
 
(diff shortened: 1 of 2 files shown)


src/solution.h
```diff
@@ -35,7 +35,11 @@ typedef enum {
 
 #define MAX_AGE_OF_DIFFERENTIAL 1.0
 
-#define DGNSS_TIMEOUT S2ST(2)
+/*Number of solution periods before timing out*/
+#define DGNSS_PERIODS 2
+#define DGNSS_TIMEOUT_MSEC (DGNSS_PERIODS * 1/((float) soln_freq) * 1000)
```
ljbade: hmm is soln_freq the name of a variable?
would it be better to make that a macro parameter?

denniszollo: It's a piksi_setting that is dynamic

denniszollo: I think that we need some protections on this setting too to avoid a user setting 0 or 10000, or a negative value.

jacobmcnamee: That should probably be a separate change, e.g. define a callback for the `soln_rate` setting and validate that.

jacobmcnamee:
I recommend moving those defines to solution.c unless they are needed externally for some reason.

Perhaps do 
```
#define DGNSS_TIMEOUT_MSEC(soln_freq_hz) (DGNSS_PERIODS * 1/((float) (soln_freq_hz)) * 1000)
```
and then pass in `soln_freq` as an argument to the #define.

denniszollo:
@jacobmcnamee this should be good to go.  Let me know what you think.

src/solution.c
```diff
@@ -42,6 +42,13 @@
 #include "signal.h"
 #include "system_monitor.h"
 
+/** number of solution periods before SPP resumes in pseudo-absolute mode */
+#define DGNSS_PERIODS 2
+
+/** number of STM clock cycles before SPP resumes in pseudo-absolute mode */
```
jacobmcnamee: STM clock cycles -> OS ticks.

src/solution.c
```diff
@@ -42,6 +42,13 @@
 #include "signal.h"
 #include "system_monitor.h"
 
+/** number of solution periods before SPP resumes in pseudo-absolute mode */
+#define DGNSS_PERIODS 2
```
jacobmcnamee: DGNSS_TIMEOUT_PERIODS ?

jacobmcnamee:
A couple minor comments, otherwise lgtm.

jacobmcnamee:
Merge when ready

PR #651, opened 2016-04-21
Final CORS fix PR 

ljbade:
Squashing into two commits to save time and get it ready for @denniszollo to start testing it ASAP.

### DIFF
--- a/src/base_obs.c
+++ b/src/base_obs.c
@@ -327,6 +327,7 @@ static void obs_callback(u16 sender_id, u8 len, u8 msg[], void* context)
       base_obss_rx.nm[base_obss_rx.n].pseudorange =
             base_obss_rx.nm[base_obss_rx.n].raw_pseudorange + clock_err * GPS_C;
       /* Set the time */
+      /* TODO: (kleeman) this is definitely wrong. */
       base_obss_rx.nm[base_obss_rx.n].tot = t;
       base_obss_rx.n++;
     }
--- a/src/base_obs.c
+++ b/src/base_obs.c
@@ -116,17 +116,55 @@ static void update_obss(obss_t *new_obss)
    * is the only thread that writes to base_obss. */
   chMtxLock(&base_obs_lock);
 
+  for (u8 i = 0; i < new_obss->n; i++) {
+    /* Set the time */
+    new_obss->nm[i].tot = new_obss->tor;
+    new_obss->nm[i].tot.tow -=
+          new_obss->nm[i].raw_pseudorange / GPS_C;
+    normalize_gps_time(&new_obss->nm[i].tot);
+
+    double clock_err;
+    double clock_rate_err;
+    /* Calculate satellite parameters using the ephemeris. */
+    ephemeris_lock();
+    ephemeris_t *e = ephemeris_get(new_obss->nm[i].sid);
+    if (calc_sat_state(e, &new_obss->nm[i].tot,
+                   new_obss->nm[i].sat_pos,
+                   new_obss->nm[i].sat_vel,
+                   &clock_err, &clock_rate_err) != 0) {
+      continue;
+    }
+    ephemeris_unlock();
+
+    /* Apply corrections to the raw pseudorange, carrier phase and Doppler. */
+    /* TODO Make a function to apply some of these corrections.
+     *      They are used in a couple places. */
+    new_obss->nm[i].pseudorange =
+          new_obss->nm[i].raw_pseudorange
+          + clock_err * GPS_C;
+    new_obss->nm[i].carrier_phase =
+          new_obss->nm[i].raw_carrier_phase
+          - clock_err * GPS_L1_HZ;
+
+    /* Used in tdcp_doppler */
+    new_obss->nm[i].doppler = clock_rate_err * GPS_L1_HZ;
+
+    /* We also apply the clock correction to the time of transmit. */
+    new_obss->nm[i].tot.tow -= clock_err;
+    normalize_gps_time(&new_obss->nm[i].tot);
+  }
+
   /* Create a set of navigation measurements to store the previous
    * observations. */
   static u8 n_old = 0;
+  static gps_time_t tor_old = {.wn = 0, .tow = 0};
   static navigation_measurement_t nm_old[MAX_CHANNELS];
 
   /* Fill in the navigation measurements in base_obss, using TDCP method to
    * calculate the Doppler shift. */
   base_obss.n = tdcp_doppler(new_obss->n, new_obss->nm,
-                             n_old, nm_old, base_obss.nm);
-  /* Copy over the time. */
-  base_obss.t = new_obss->t;
+                             n_old, nm_old, base_obss.nm,
+                             gpsdifftime(&new_obss->tor, &tor_old));
 
   /* Copy over sender ID. */
   base_obss.sender_id = new_obss->sender_id;
@@ -136,23 +174,16 @@ static void update_obss(obss_t *new_obss)
   memcpy(nm_old, new_obss->nm,
          new_obss->n * sizeof(navigation_measurement_t));
   n_old = new_obss->n;
-  
+  tor_old = new_obss->tor;
+
+  /* Copy over the time. */
+  base_obss.tor = new_obss->tor;
+
   /* Reset the `has_pos` flag. */
   u8 has_pos_old = base_obss.has_pos;
   base_obss.has_pos = 0;
-  /* Check if the base station has sent us its position explicitly via a
-   * BASE_POS SBP message (as indicated by #base_pos_known), and if so use
-   * that. No need to lock before reading here as base_pos_* is only written
-   * from this thread (SBP).
-   */
-  if (base_pos_known) {
-    /* Copy the known base station position into `base_obss`. */
-    memcpy(base_obss.pos_ecef, base_pos_ecef, sizeof(base_pos_ecef));
-    /* Indicate that the position is valid. */
-    base_obss.has_pos = 1;
-  /* The base station wasn't sent to us explicitly but if we have >= 4
-   * satellites we can calculate it ourselves (approximately). */
-  } else if (base_obss.n >= 4) {
+
+  if (base_obss.n >= 4) {
     gnss_solution soln;
     dops_t dops;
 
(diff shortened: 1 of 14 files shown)


ljbade:
Need to check has POS flag for race condition
Also push hitl tag

libswiftnav
```diff
@@ -1 +1 @@
-Subproject commit 45657c98c5caba708a7c3116a7e77d4f30f04054
+Subproject commit 428ac469b11f63d9ba71d8c0b1e24be22f527b0e
```
jacobmcnamee: remove libswiftnav update

src/solution.c
```diff
@@ -409,15 +410,34 @@ static void solution_thread(void *arg)
...
 
     ephemeris_lock();
-    calc_navigation_measurement(n_ready, p_meas, p_nav_meas,
-                                (double)((u32)nav_tc)/SAMPLE_FREQ, p_e_meas);
+    gps_time_t nav_time = rx2gpstime(nav_tc);
+    if (time_quality == TIME_FINE) {
+      /* If we have timing then we can calculate the relationship between
+       * receiver time and GPS time and hence provide the pseudorange
+       * calculation with the local GPS time of reception. */
+      calc_navigation_measurement(n_ready, p_meas, p_nav_meas,
```
jacobmcnamee: Conditionally define the `nav_time` pointer instead of duplicating everything:
`p_nav_time = (time_quality == TIME_FINE) ? &nav_time : NULL;`

src/solution.c
```diff
@@ -440,6 +440,11 @@ static void solution_thread(void *arg)
     }
     ephemeris_unlock();
 
+    for (u32 i=0; i<n_ready; i++) {
+      printf("%d: tot: %.7f rpr: %.7f", nav_meas[i].sid.sat, nav_meas[i].tot.tow,
```
jacobmcnamee: remove `printf()`

jacobmcnamee:
Please remove the libswiftnav pointer updates from intermediate commits. Only include valid pointers to libswiftnav/master.

src/base_obs.h
```diff
@@ -45,6 +45,11 @@ typedef struct {
 /** Maximum difference between observation times to consider them matched. */
 #define TIME_MATCH_THRESHOLD 2e-3
 
+/* Maximum distance between calculated and surveyed base station single point
+ * position for error checking.
+ */
+#define BASE_STATION_DISTANCE_THRESHOLD 5
```
jacobmcnamee: units?

src/board/v2/error.c
```diff
@@ -182,7 +182,7 @@ void HardFaultVector(void)
   static uint32_t *psp, *lr;
   asm("mrs %0, psp" : "=r"(psp) : : );
   asm("mov %0, lr" : "=r"(lr) : : );
-  
+
```
jacobmcnamee: remove

src/solution.c
```diff
@@ -403,56 +402,53 @@ static void solution_thread(void *arg)
...
-    gps_time_t nav_time = rx2gpstime(nav_tc);
     if (time_quality == TIME_FINE) {
       /* If we have timing then we can calculate the relationship between
        * receiver time and GPS time and hence provide the pseudorange
        * calculation with the local GPS time of reception. */
       calc_navigation_measurement(n_ready, p_meas, p_nav_meas,
-                                  (double)((u32)nav_tc)/SAMPLE_FREQ,
-								  &nav_time, p_e_meas);
+                                  (double)((u32)rec_tc) / SAMPLE_FREQ,
+								                  &rec_time, p_e_meas);
```
jacobmcnamee: tabs?

src/solution.c
```diff
@@ -512,46 +515,58 @@ static void solution_thread(void *arg)
...
+     *  The last bit comes from TOT(tor) == TOT(t_pvt), which simply stems from
+     *  the fact that uncorrected and corrected pseudoranges correspond to the
+     *  exact same observations.
+     */
+    for (u8 i = 0; i < n_ready_tdcp; i++) {
+      nav_meas_tdcp[i].raw_pseudorange += GPS_C * gpsdifftime(&position_solution.time, &rec_time);
+      nav_meas_tdcp[i].pseudorange += GPS_C * gpsdifftime(&position_solution.time, &rec_time);
     }
-    chMtxUnlock(&base_obs_lock);
+    /**
```
jacobmcnamee: /*

src/solution.c
```diff
@@ -577,8 +577,7 @@ static void solution_thread(void *arg)
       /* Propagate observation to desired time. */
       /* We have to use the tdcp_doppler result to account for TCXO drift. */
       for (u8 i = 0; i < n_ready_tdcp; i++) {
-        nav_meas_tdcp[i].raw_pseudorange -= t_err * nav_meas_tdcp[i].raw_doppler
-                                            * GPS_L1_LAMBDA;
+        nav_meas_tdcp[i].raw_pseudorange -= t_err;
```
jacobmcnamee: squash this change?

src/track.c
```diff
@@ -616,6 +624,37 @@ void tracking_channel_measurement_get(tracker_channel_id_t id,
...
   meas->lock_counter = internal_data->lock_counter;
+
+  /* Check if lock counter changed */
+  if (tracker_channel->carrier_phase_offset_lock
+      != internal_data->lock_counter) {
+    tracker_channel->carrier_phase_offset_lock = internal_data->lock_counter;
+    tracker_channel->carrier_phase_offset = 0.0;
+  }
+
+  /* Adjust carrier phase initial integer offset to be approximately equal to
```
jacobmcnamee: Remove commented code or combine with commit that uncomments it

src/track.c
```diff
@@ -344,6 +349,9 @@ bool tracker_channel_init(tracker_channel_id_t id, gnss_signal_t sid,
 
     tracker_channel->elevation = elevation;
 
+    /* Reset the carrier phase offset so it gets set again. */
+    tracker_channel->carrier_phase_offset = 0.0;
```
jacobmcnamee: also set `carrier_phase_offset_lock` ?

src/nmea.c
```diff
@@ -393,22 +393,26 @@ void nmea_gpgll(const gnss_solution *soln, const gps_time_t *gps_t)
...
   DO_EVERY(gpgsv_msg_rate,
     nmea_gpgsv(n, nm, soln);
   );
-  DO_EVERY(gprmc_msg_rate,
-    nmea_gprmc(soln, &soln->time);
-  );
-  DO_EVERY(gpvtg_msg_rate,
-    nmea_gpvtg(soln);
-  );
+  if (!skip_velocity) {
```
jacobmcnamee: @denniszollo Is this going to be a problem?

jacobmcnamee: Should we just skip all solution outputs if this happens?

src/solution.c
```diff
@@ -602,8 +599,15 @@ static void solution_thread(void *arg)
         ephemeris_unlock();
       }
 
-      /* Send the observations. */
```
jacobmcnamee: It looks like this commit is supposed to 
1. still propagate and post back to the mailbox even if `fabs(t_check - (u32)t_check) < TIME_MATCH_THRESHOLD)` fails
2. additionally check for `TIME_FINE` before sending observations

Am I missing something, or is the commit message inaccurate?

src/base_obs.c
```diff
@@ -174,10 +174,7 @@ static void update_obss(obss_t *new_obss)
   /* Copy over the time. */
   base_obss.tor = new_obss->tor;
 
-  /* Reset the `has_pos` flag. */
```
jacobmcnamee: I don't think the change is necessary as all accesses to `base_obss` should be mutex protected.

jacobmcnamee:
FYI I went through commit by commit, so there are some "comments on outdated diffs"

ljbade:
@jacobmcnamee Thanks, ill get to work on them, and the rebase.

src/solution.c
ljbade: This was removed in a subsequent commit.

jacobmcnamee:
Instead of duplicating the lock counter, just clear the carrier phase offset in `tracker_ambiguity_unknown()` (http://github.com/swift-nav/piksi_firmware/blob/master/src/track_api.c#L221)

src/board/v2/error.c
ljbade: Is there supposed to be a space here? My editor must have automatically removed when I was looking at this file...

src/solution.c
ljbade: Already fixed in subsequent commit.

src/solution.c
ljbade: I made this a separate commit to keep track of who came up with this change.

src/nmea.c
ljbade: When I chatted with @denniszollo last week his preference was to just skip the messages with a velocity..

src/base_obs.c
ljbade: Ah, OK @denniszollo spotted this when we were figuring out the 5Hz stuff.

@denniszollo What is your preference here?

src/solution.c
ljbade: This confusing due to the way the silly `(u32)t_check` thing is used in firmware from before this PR.

The default settings for Piksi is that `soln_freq` is 10Hz, and `obs_output_divisor` is 2 to give effective rate of 5Hz. Thus you will get a position solution in SBP every 10Hz, and a observation message every 5Hz.

Due to some rearranging of the solution thread code in this PR, the 5Hz check was also covering the position solution code.

So now the time propagation (which is needed for the position solution), and the solution code are run at 5Hz, but the code that sends to observations only runs at 5Hz, and only when the receiver clock has been correctly set (`TIME_FINE`)

jacobmcnamee:
The sample count for each channel is now managed in the NAP code and simply copied out to the common data struct at each update. See `nap_track_read_results()`. For v2, it would be sufficient to make the sample count (actually called `count_snapshot` now) a u64 as you did in track.c previously. But for v3 the sample count is read out of the NAP each update and not currently stored in software, which is problematic since that value is always a u32 in the NAP (whereas you need the u64 software-extended version).

ljbade:
@jacobmcnamee I have rebased on your https://github.com/swift-nav/piksi_firmware/pull/674

I have also used `tracker_ambiguity_unknown()` instead of the lock counter, and moved the offset to the internal struct.

Please review.

src/manage.c
```diff
@@ -678,6 +678,18 @@ static void manage_tracking_startup(void)
       continue;
     }
 
+    /* Transition to tracking. */
```
jacobmcnamee: Remove this. The u64 timing count stuff is unnecessary with the measurement time delta changes.

src/system_monitor.c
```diff
@@ -177,11 +173,10 @@ static void system_monitor_thread(void *arg)
       llhdeg2rad(base_llh, tmp);
       wgsllh2ecef(tmp, base_ecef);
 
-      vector_subtract(3, base_ecef, position_solution.pos_ecef, tmp);
-      base_distance = vector_norm(3, tmp);
+      base_distance = vector_distance(3, base_ecef, position_solution.pos_ecef);
 
       if (base_distance > BASE_STATION_DISTANCE_THRESHOLD) {
-        log_warn("Invalid surveyed position coordinates\n");
+        log_warn("Sending invalid surveyed position coordinates.");
```
jacobmcnamee: They won't actually be sent if this is printed, correct?

src/track.h
```diff
@@ -47,7 +47,7 @@ void tracking_channels_missed_update_error(u32 channels_mask);
 /* State management interface */
 bool tracker_channel_available(tracker_channel_id_t id, gnss_signal_t sid);
 bool tracker_channel_init(tracker_channel_id_t id, gnss_signal_t sid,
-                          u32 start_sample_count, float code_phase,
```
jacobmcnamee: remove

src/system_monitor.c
denniszollo: affirmative.  Perhaps make it this: "Invalid surveyed position coordinates. No base position msg will be sent."

src/manage.c
ljbade: Hmm this might have been an artifact of the rebase

src/track.h
ljbade: I noticed this, it didnt match the variable name in track.c's `tracker_channel_init`

jacobmcnamee: Hmm ok. In that case blame Gareth.

ljbade:
This PR now depends on https://github.com/swift-nav/piksi_firmware/pull/674

ljbade:
Merging once travis passes

PR #652, opened 2016-04-22
Require the solution thread to sleep for a minimum amount of time

jacobmcnamee:
This will cause solution epochs to be skipped if the solution thread is using too much CPU time. Mitigates issue where decode / manage threads are starved and a watchdog reset is triggered.


gsmcmullin:
lgtm

PR #653, opened 2016-04-22
Call ephemeris_setup() earlier

jacobmcnamee:
Ephemeris is in CCM and not cleared to zero at startup (before call to ephemeris_setup()). Garbage ephemeris can thus appear to be valid if the valid field is nonzero.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -145,6 +145,7 @@ int main(void)
   init();
   settings_setup();
   signal_init();
+  ephemeris_setup();
 
   check_nap_auth();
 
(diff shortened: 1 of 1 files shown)


gsmcmullin:
What is consuming the ephemeris too early that this fixes?

jacobmcnamee:
- Solution thread calls `use_tracking_channel()` which checks ephemeris
- Base obs (via SBP) checks ephemeris for incoming messages
There are certainly others which are less likely.

gsmcmullin:
Go ahead and merge.

PR #655, opened 2016-04-29
v3: Get version and device DNA from NAPv3.

wltr:
Reads out the version and device DNA from SwiftNAPv3 and provides a function to unlock the NAP when AES authentication is used.

/cc @jacobmcnamee @gsmcmullin


src/board/v3/nap/nap_constants.h
```diff
@@ -19,4 +19,6 @@
 #define NAP_ACQ_SAMPLE_RATE_Hz                  (NAP_FRONTEND_SAMPLE_RATE_Hz  \
                                                    / NAP_ACQ_DECIMATION_RATE)
 
+#define NAP_DNA_OFFSET                                                   (14)
```
jacobmcnamee: Can you define offset and length for DNA and version?

src/board/v3/nap/nap_common.c
```diff
@@ -119,3 +124,55 @@ static void nap_exti_thread(void *arg)
   }
 }
 
+u8 nap_conf_rd_version_string(char version_string[])
+{
+  u8 i = 0;
+  u32 reg = 0;
+
+  do {
+    NAP->CONTROL = (i << 1);
```
jacobmcnamee: I like Gareth's suggestion of keeping the method of reading the version constant across all HDL versions so that we can gracefully warn instead of stalling the bus. Is that reasonable with this method / any changes you might want to make before committing to never changing it?

jacobmcnamee: Also, is there anything else in the config register besides the offset for DNA / version?

src/board/v3/nap/nap_common.c
```diff
@@ -119,3 +124,55 @@ static void nap_exti_thread(void *arg)
...
+  static sbp_msg_callbacks_node_t nap_dna_node;
+
+  sbp_register_cbk(SBP_MSG_NAP_DEVICE_DNA_REQ, &nap_rd_dna_callback,
+      &nap_dna_node);
+}
+
+void nap_unlock(const u8 key[])
+{
+  for (u8 i = 0; i < 16; ++i) {
+    NAP->CONTROL = ((u32)key[i] << 24) | (i << 20);
```
jacobmcnamee: Define offset for these in nap_hw.h

src/board/v3/nap/nap_common.c
wltr: Yes, two other bits. One to select the raw sample grabber input and the other for selecting external event edge sensitivity. I assumed we would only read out the version at startup and the DNA during device bringup so it wouldn't matter if you overwrite anything. But probably better to play it safe. I'll change that.

src/board/v3/nap/nap_common.c
```diff
@@ -119,3 +124,64 @@ static void nap_exti_thread(void *arg)
   }
 }
 
+u32 nap_conf_rd_version(void)
```
wltr: @jacobmcnamee: `nap_conf_rd_version()` should be function to always check the version. format: major (1 byte), minor (1 byte), build (2 bytes). major represents NAPv3, I'll update minor whenever the register interface changes, and build will be representative of smaller features.

src/board/v3/nap/nap_common.c
```diff
@@ -119,3 +124,64 @@ static void nap_exti_thread(void *arg)
...
+    reg = NAP->VERSION;
+    memcpy(dna + i * sizeof(reg), &reg, sizeof(reg));
+  }
+}
+
+void nap_rd_dna_callback(u16 sender_id, u8 len, u8 msg[], void* context)
+{
+  (void)sender_id; (void)len; (void)msg; (void) context;
+  u8 dna[8];
+  nap_rd_dna(dna);
```
jacobmcnamee: This is not thread-safe due to the writes to `NAP->CONTROL`. Can we copy the DNA to a static variable once and use that instead?

wltr: Yes, I'll change that.

src/board/v3/nap/nap_common.c
```diff
@@ -164,10 +166,7 @@ void nap_rd_dna(u8 dna[])
 void nap_rd_dna_callback(u16 sender_id, u8 len, u8 msg[], void* context)
 {
   (void)sender_id; (void)len; (void)msg; (void) context;
-  u8 dna[8];
-  nap_rd_dna(dna);
-
-  sbp_send_msg(SBP_MSG_NAP_DEVICE_DNA_RESP, 8, dna);
+  sbp_send_msg(SBP_MSG_NAP_DEVICE_DNA_RESP, 8, nap_dna);
```
wltr: @jacobmcnamee better?

src/board/v3/nap/nap_common.c
```diff
@@ -119,3 +126,61 @@ static void nap_exti_thread(void *arg)
...
+        << NAP_CONTROL_VERSION_ADDR_Pos);
+    reg = NAP->VERSION;
+    memcpy(dna + i * sizeof(reg), &reg, sizeof(reg));
+  }
+}
+
+void nap_rd_dna_callback(u16 sender_id, u8 len, u8 msg[], void* context)
+{
+  (void)sender_id; (void)len; (void)msg; (void) context;
+  sbp_send_msg(SBP_MSG_NAP_DEVICE_DNA_RESP, 8, nap_dna);
```
wltr: @jacobmcnamee better?

src/board/v3/nap/nap_common.c
```diff
@@ -21,19 +22,25 @@
...
 #define PROCESS_PERIOD_ms (1000)
 
 void nap_isr(void *context);
+void nap_unlock(const u8 key[]);
 
 static BSEMAPHORE_DECL(nap_exti_sem, TRUE);
 static WORKING_AREA_CCM(wa_nap_exti, 2000);
 static void nap_exti_thread(void *arg);
 
+static u8 nap_dna[8] = {0};
```
jacobmcnamee: `nap_dna[NAP_DNA_LENGTH]`

wltr: `NAP_DNA_LENGTH` is actually 2 as it's 2 x u32 in the NAP ROM. I changed it to:
`u8 nap_dna[NAP_DNA_LENGTH * sizeof(u32)]`

jacobmcnamee: Are the other offsets / lengths in bytes? Seems more consistent to use bytes for everything.

jacobmcnamee:
Minor: Can you get rid of the pointer arithmetic e.g. `memcpy(dna + i * sizeof(reg), &reg, sizeof(reg))`. The same when reading version. `&dna[i * sizeof(reg)]`

wltr:
Now all offsets and lengths are in bytes and I removed the pointer arithmetic.

src/board/v3/nap/nap_common.c
```diff
@@ -119,3 +126,61 @@ static void nap_exti_thread(void *arg)
...
+  static sbp_msg_callbacks_node_t nap_dna_node;
+
+  sbp_register_cbk(SBP_MSG_NAP_DEVICE_DNA_REQ, &nap_rd_dna_callback,
+      &nap_dna_node);
+}
+
+void nap_unlock(const u8 key[])
+{
+  for (u8 i = 0; i < NAP_KEY_LENGTH; ++i) {
+    NAP->CONTROL = ((u32)key[i] << NAP_CONTROL_KEY_BYTE_Pos) |
```
jacobmcnamee: Can you do the same masking performed in the other functions here?

src/board/v3/nap/nap_common.c
jacobmcnamee: Great.

jacobmcnamee:
One comment, otherwise lgtm.

src/board/v3/nap/nap_common.c
```diff
@@ -119,3 +126,64 @@ static void nap_exti_thread(void *arg)
...
+void nap_unlock(const u8 key[])
+{
+  u32 ctrl = (NAP->CONTROL & ~((u32)NAP_CONTROL_KEY_ADDR_Msk |
+        NAP_CONTROL_KEY_BYTE_Msk));
+
+  for (u8 i = 0; i < NAP_KEY_LENGTH; ++i) {
+    NAP->CONTROL = ctrl | ((u32)key[i] << NAP_CONTROL_KEY_BYTE_Pos) |
+        (i << NAP_CONTROL_KEY_ADDR_Pos);
+  }
+}
```
wltr: @jacobmcnamee there you go

wltr:
Tested on uZed with different NAP versions. Merging.

PR #657, opened 2016-04-29
NMEA: Removed terminating NULL character of output

jkretzmer:
Previously the NMEA output incorrectly terminated with a NULL. According to the spec it should only be <CR><LF> - this was causing the Skydev simulator choke on our NMEA stream. I made a change to simply not output the last character over the serial, as I wasn't sure if other parts of the source base rely on the NMEA sentences being NULL terminated.

This change was built and tested against the Skydev simulator which now is able to read the stream correctly.

CC: @jacobmcnamee @fnoble @switanis 

This should be reviewed.

You can find the NMEA standard here: https://drive.google.com/open?id=0B51x3WASpitCSGoyRGVLdVBHWHM

Section 5.3.1 is relevant.

### DIFF
--- a/src/nmea.c
+++ b/src/nmea.c
@@ -65,18 +65,25 @@ static struct nmea_dispatcher *nmea_dispatchers_head;
     if (sentence_bufp >= sentence_buf_end) \
       sentence_bufp = sentence_buf_end; } while (0)
 
-/** NMEA_SENTENCE_DONE: append checksum and dispatch. */
+/** NMEA_SENTENCE_DONE: append checksum and dispatch. 
+ * \note According to section 5.3.1 of the NMEA 0183 spec, sentences are
+ *       terminated with <CR><LF>. The sentence_buf is null_terminated.
+ *       The call to nmea_output has been modified to remove the NULL.
+ *       This will also affect all registered dispatchers
+ */
 #define NMEA_SENTENCE_DONE() do { \
     if (sentence_bufp == sentence_buf_end) \
       log_warn("NMEA %.6s cut off", sentence_buf); \
     nmea_append_checksum(sentence_buf, sizeof(sentence_buf)); \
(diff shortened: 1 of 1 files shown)


jacobmcnamee:
Most standard library functions exclude the terminating null when computing length, so I think it would be more consistent and cleaner to leave `nmea_output()` unchanged and instead call that function with `length-1`. I.e. just modify the `NMEA_SENTENCE_DONE` macro.

jkretzmer:
@jacobmcnamee i updated the macro as per your suggestion and left nmea_output as it was. please let me know if there's anything else that needs to be addressed. thanks!

src/nmea.c
```diff
@@ -65,25 +65,33 @@ static struct nmea_dispatcher *nmea_dispatchers_head;
...
  * \param s The NMEA sentence to output.
+ * \param size This is the C-string size, not including the null character
  */
 static void nmea_output(char *s, size_t size)
 {
   static MUTEX_DECL(send_mutex);
   chMtxLock(&send_mutex);
 
   if ((ftdi_usart.mode == NMEA) && usart_claim(&ftdi_state, NMEA_MODULE)) {
+
```
jacobmcnamee: remove newline

jacobmcnamee:
@denniszollo Is there anything you want to do to test this?

denniszollo:
@switanis Could you run the output of this firmware through your NMEA validator as to verify that we've fixed the null terminator oddity?

denniszollo:
We will just do a hitl run and verify no crashes.  I've tested on the bench.

jkretzmer:
HITL results: http://sbp-log-analysis.swiftnav.com/#/d/0/q/x/scenario/y/metric/f/metric/p/PRDv3/f/firmware/sv/piksi_firmware_v0.21-192-ga0e453f.hex?_k=1iv2kr

denniszollo:
No crashes over 16 runs and nothing out of the ordinary - LGTM

jacobmcnamee:
Remove the extra newline and then lgtm

PR #658, opened 2016-05-02
Fix thread state names

rge-exafore:


rge-exafore:
@swift-nav/firmware please review

jacobmcnamee:
lgtm

PR #659, opened 2016-05-02
usart v2: Disable RX DMA and reset state when reconfiguring.

gsmcmullin:
This fixes the DMA RX buffer overrun message when changing baudrates.

/cc @swift-nav/firmware


jacobmcnamee:
I tested manually and was unable to reproduce the issue with this change. I vote we merge since HITL won't really exercise any of the affected code.

jacobmcnamee:
@gsmcmullin Merge if you're happy with it.

PR #661, opened 2016-05-03
Update the OSX setup.sh install script

margaret:
- Need to install the GCC ARM embedded before Ansible

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -113,6 +113,13 @@ function bootstrap_osx () {
         log_info "Installing homebrew..."
         homebrew_install
     fi
+    # Download and install GCC ARM Embedded
+    if [[ ! -x /usr/local/Cellar/gcc-arm-none-eabi/ ]]; then
+        log_info "Installing gcc-arm-none-eabi..."
+        brew tap PX4/homebrew-px4
+        brew update
+        brew install gcc-arm-none-eabi
+    fi
     # Download and install Ansible
     if [[ ! -x /usr/local/bin/ansible ]]; then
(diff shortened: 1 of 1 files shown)


mfine:
Hmm, I think all that ansible is doing now is installing GCC arm :) We should purge the ansible stuff and move the arm setup things into the setup.sh script, as ansible seems like total overkill now. (Can make it a issue for future work).

jacobmcnamee:
What is `brew tap PX4/homebrew-px4` for?

mookerji:
👍 I wrote the ansible stuff my second day at Swift and we can totally remove it. I agree let's do it with a separate PR or issue. 

I will note additionally that we do rely on a fixed version of the gcc arm stuff and sometimes having the wrong version of it can cause the firmware to fail to compile. Also, depending on vendor'd dependencies from things like homebrew or unstable Debian is also problematic, since the maintainers of those repositories will sometimes disappear. I would recommend just mirroring the gcc tarball we need to s3 and then manually installing it from there using the make tools.

margaret:
This one seems to have installed 

```
$ brew install gcc-arm-none-eabi
==> Installing gcc-arm-none-eabi from px4/px4
==> Downloading http://launchpadlibrarian.net/186124092/gcc-arm-none-eabi-4_8-2014q3-20140805-mac.tar.bz2
```
Currently seeing if the build works with it. @mookerji Do you know what the version ranges it works with is?

@jacobmcnamee `brew tap [repo]` tells homebrew where to get the package

PR #662, opened 2016-05-04
cn0: added separation of C/N0 computation and filtering

valeri-atamaniouk:
Added separation of C/N0 computation and value filtering.

Depends on https://github.com/swift-nav/libswiftnav/pull/341


valeri-atamaniouk:
Work moved into private

PR #665, opened 2016-05-05
Fault handling improvements

jacobmcnamee:
v3: Switch to SYS stack and log LR, e.g.
`ERROR: /Swift/piksi_firmware//src/board/v3/error.c:140 : Prefetch Abort! lr=00110E7A`
Should be enough to figure out what instruction caused the problem.

v2: Update handler names to match symbols expected by ChibiOS 3

general: Bypass thread-safe `sprintf()` in fault handlers

### DIFF
--- a/src/board/v3/error.c
+++ b/src/board/v3/error.c
@@ -122,7 +122,8 @@ void fault_handling_setup(void) {
 void fault_handler_screaming_death(const char *msg_str, u32 lr)
 {
   char msg[128];
-  sprintf(msg, "%s lr=%08X", msg_str, (unsigned int)lr);
+  extern int fallback_sprintf(char *str, const char *fmt, ...);
+  fallback_sprintf(msg, "%s lr=%08X", msg_str, (unsigned int)lr);
   screaming_death(msg);
 }
 
--- a/src/board/v3/error.c
+++ b/src/board/v3/error.c
@@ -71,7 +71,7 @@ void _screaming_death(const char *pos, const char *msg)
   sbp_state_init(&sbp_state);
 
   /* Continuously send error message */
-  #define APPROX_ONE_SEC 33000000
+  #define APPROX_ONE_SEC 200000000
   while (1) {
     led_toggle(LED_RED);
     for (u32 d = 0; d < APPROX_ONE_SEC; d++) {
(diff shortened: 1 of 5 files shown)


src/board/v3/error.c
```diff
@@ -30,6 +30,16 @@
  * System low-level error handling and reporting
  * \{ */
 
+#define FAULT_HANDLER_SCREAMING_DEATH(msg_str)                                \
+do {                                                                          \
+  uint32_t *lr;                                                               \
+  asm("mov %0, lr" : "=r"(lr) : : );                                          \
```
gsmcmullin: I'd prefer `register uint32_t *lr asm ("lr");`

src/board/v3/error.c
```diff
@@ -30,6 +30,16 @@
...
+  char msg[128];                                                              \
```
gsmcmullin: These fault handlers have super small stacks.  We should probably switch to another CPU mode before allocating a big buffer or calling sprintf.

src/board/v3/error.c
```diff
@@ -30,6 +30,16 @@
  * System low-level error handling and reporting
  * \{ */
 
+#define FAULT_HANDLER_SCREAMING_DEATH(msg_str)                                \
```
gsmcmullin: I'd prefer a function

src/board/v3/error.c
```diff
@@ -71,7 +81,7 @@ void _screaming_death(const char *pos, const char *msg)
   sbp_state_init(&sbp_state);
 
   /* Continuously send error message */
-  #define APPROX_ONE_SEC 33000000
+  #define APPROX_ONE_SEC 200000000
```
gsmcmullin: This was always terrible.  We should use the cycle counter for this instead.

src/board/v3/error.c
jacobmcnamee: Right, but you can't get the `LR` that we want that way. You'd just get the address of the fault handler itself.

gsmcmullin: Good point

src/board/v3/error.c
jacobmcnamee: I forget the exact reason, but we're actually using one of the timer peripherals as the cycle counter for ChibiOS on v3. So for purposes of being failsafe, I think this should stay the way it is (though I agree it's terrible).

src/board/v3/error.c
jacobmcnamee: Working on this.

PR #668, opened 2016-05-10
Fix assert due to watchdog_notify() before system_monitor_setup()

jacobmcnamee:
Also gets rid of the chEvtSignal() stuff


gsmcmullin:
lgtm

PR #672, opened 2016-05-12
Add L2C tracking

adel-mamin:
This PR add L2C tracking capability on Piksi V3 HW. It is contingent on PR:
https://github.com/swift-nav/libswiftnav-private/pull/6


src/board/v3/nap/track_channel.c
```diff
@@ -50,45 +50,177 @@ static struct {
...
+                 chips_to_correlate - cp_start;
   u32 samples = cp_units / cp_rate;
   return samples;
 }
 
+/** Return the init value of NAP G1 pseudorandom generator.
+ * \param sid Signal ID
+ * \return NAP G1 initial value
+ */
+u32 sid_to_init_g1(gnss_signal_t sid)
```
jacobmcnamee: Can this function be moved to libswiftnav?

src/board/v3/nap/track_channel.c
```diff
@@ -50,45 +50,177 @@ static struct {
...
+  }
+
+  return ret;
+}
+
+/** Looks-up RF frontend channel for the given signal ID.
+ * \param sid Signal ID.
+ * \return RF front-end channel number.
+ */
+u8 sid_to_rf_frontend_channel(gnss_signal_t sid)
```
jacobmcnamee: Fine for now, but eventually these functions will need to be shared with acquisition.

src/sbp_utils.c
```diff
@@ -294,7 +294,7 @@ void unpack_ephemeris(const msg_ephemeris_t *msg, ephemeris_t *e)
   e->sid              = sid_from_sbp(msg->sid);
   e->kepler.iode      = msg->iode;
   e->kepler.iodc      = msg->iodc;
-  e->fit_interval     = 4 * 60 * 60; /* TODO: this is a work around until SBP updated */
+  e->fit_interval     = (u8)(4 * 60 * 60); /* TODO: this is a work around until SBP updated */
```
jacobmcnamee: ? This is going to overflow. Is it actually a u8?

src/track/track_gps_l2cm.c
```diff
@@ -0,0 +1,638 @@
...
+  l2c_cpbl |= (u32)1 << (25 - 1);
+  l2c_cpbl |= (u32)1 << (26 - 1);
+  l2c_cpbl |= (u32)1 << (27 - 1);
+  l2c_cpbl |= (u32)1 << (29 - 1);
+  l2c_cpbl |= (u32)1 << (30 - 1);
+  l2c_cpbl |= (u32)1 << (31 - 1);
+  l2c_cpbl |= (u32)1 << (32 - 1);
+
+  /* compose SID: same SV, but code is L2 CM */
+  gnss_signal_t sid = {
```
jacobmcnamee: Use `construct_sid()`

src/track/track_gps_l2cm.c
```diff
@@ -0,0 +1,638 @@
...
+                 code_phase);
+    return;
+  }
+
+  if (code_phase > (GPS_L1CA_CHIPS_NUM - 0.5)) {
+    code_phase = 2 * GPS_L2CM_CHIPS_NUM - (GPS_L1CA_CHIPS_NUM - code_phase);
+  }
+
+  /* recalculate doppler freq for L2 from L1*/
+  double carrier_freq = tracking_channel_carrier_freq_get(nap_channel) *
```
jacobmcnamee: I recommend you pass in the doppler frequency and cn0 instead of calling `tracking_channel_get()` functions. For elevation just use `TRACKING_ELEVATION_UNKNOWN`.

src/track_api.c
```diff
@@ -156,6 +156,13 @@ void tracker_bit_sync_update(tracker_context_t *context, u32 int_ms,
   tracker_internal_data_t *internal_data;
   tracker_internal_context_resolve(context, &channel_info, &internal_data);
 
+  /* L2C bit sync is known once we start tracking it since
```
jacobmcnamee: It would be cleaner and more flexible to add a `bit_sync_set()` function in bit_sync.c (libswiftnav) which simply sets `bit_phase_ref` to the specified value. Then add a function to this file like `tracker_bit_sync_set()` which calls `bit_sync_set()`.

With those changes you should be able to leave `tracker_bit_sync_update()` unchanged and get nav bit output.

src/track/track_gps_l2cm.c
```diff
@@ -0,0 +1,638 @@
...
+  tracker_interface_register(&tracker_interface_list_element_gps_l2cm);
+}
+
+/** Check if a satellite is already in track
+ *
+ * \param sat Satellite ID
+ * \retval false Satellite is not in track
+ * \retval true Satellite is in track
+ */
+static bool is_in_track(u16 sat)
```
jacobmcnamee: Let's move this to manage.c:
```
bool tracking_startup_ready(gnss_signal_t sid)
{
  u16 global_index = sid_to_global_index(sid);
  acq_status_t *acq = &acq_status[global_index];
  return (acq->state == ACQ_PRN_ACQUIRING);
}
```

jacobmcnamee: Also, now that we're adding tracking startup requests from multiple threads, I realized we need an additional check in `manage_tracking_startup()`:
```
     acq_status_t *acq = &acq_status[sid_to_global_index(startup_params.sid)];
 
+    /* Make sure the SID is not already tracked. */
+    if (acq->state == ACQ_PRN_TRACKING) {
+      continue;
+    }
+
     /* Make sure a tracking channel and a decoder channel are available */
     u8 chan = manage_track_new_acq(startup_params.sid);
```
If you want to copy that into your PR that would be great :)

jacobmcnamee:
@gsmcmullin Can you take a look at the `chips_to_correlate` stuff?

src/board/v3/nap/track_channel.c
adel-mamin: This function is specific to NAP v3. As this file is in v3/nap directory, it looks to me that it belongs here quite naturally. I could not find a better place for this function in libswiftnav. Anyways, if you could pinpoint the place in libswiftnav where I could move this function, then I will move it there.

src/board/v3/nap/track_channel.c
adel-mamin: Ok. I see.

src/sbp_utils.c
adel-mamin: My bad. This was needed to pass compilation stage before I rebased to piksi_firmware/master. It is not needed anymore. I will remove this code change.

src/track/track_gps_l2cm.c
adel-mamin: I will fix it.

src/track/track_gps_l2cm.c
adel-mamin: I will replace elevation with TRACKING_ELEVATION_UNKNOWN assuming that manage_tracking_startup() will take care of it.
I am a bit confused about your comment regarding the doppler and cn0. If we know the doppler on L1C/A, it cannot be used as is for L2C tracker initialization because L2C has a different carrier frequency. So L1C/A doppler value must be scaled and the scaling factor is GPS_L2_HZ / GPS_L1_HZ. That is what the code is doing here. CN0 init value is taken to be the same as CN0 of L1C/A as the best estimation of CN0 at this stage.
So could you please elaborate?

src/track_api.c
adel-mamin: I will fix it.

src/track/track_gps_l2cm.c
adel-mamin: Should be done now.

src/board/v3/nap/track_channel.c
jacobmcnamee: At least the data from the ICD can be moved to libswiftnav: `The L2C G1 init values are taken from IS-GPS-200H Table 3-IIa`. Probably `prns.c`

src/track/track_gps_l2cm.c
jacobmcnamee: I mean pass the L1CA doppler and cn0 in as arguments to `do_l1ca_to_l2cm_handover()`. That function is called from `tracker_gps_l1ca_update()`, where you can access the L1CA parameters as `common_data->carrier_freq` and `common_data->cn0`.

src/track/track_gps_l2cm.c
```diff
@@ -0,0 +1,614 @@
...
+  }
+
+  /* recalculate doppler freq for L2 from L1*/
+  double carrier_freq = tracking_channel_carrier_freq_get(nap_channel) *
+                        GPS_L2_HZ / GPS_L1_HZ;
+
+  /* get initial cn0 from parent L1 channel */
+  float cn0_init = tracking_channel_cn0_get(nap_channel);
+
+  /* The best elevation estimation could be retrieved by calling
```
jacobmcnamee: FYI I think elevation should go in the NDB or somewhere other than the tracking channels eventually. Right now it is computed by the solution thread periodically and written back to the tracking channels, which which is clumsy. I'm mostly just trying to maintain the architecture where `track_gps_l1ca.c` and similar are slaves to `track.c` and only access data via the API in `track_api.h`

src/track/track_gps_l2cm.c
```diff
@@ -0,0 +1,614 @@
...
+                                    tracker_common_data_t *common_data,
+                                    tracker_data_t *tracker_data)
+{
+  gps_l2cm_tracker_data_t *data = tracker_data;
+
+  if (data->startup) {
+    /* Handle start-up case, when we have 1+1+18 integration cycles */
+    handle_tracker_startup(channel_info, common_data, data);
+    if (1 == data->startup) {
+      // The start-up phase is over. Continue with the normal
```
jacobmcnamee: /* */

src/track/track_gps_l2cm.c
```diff
@@ -0,0 +1,614 @@
...
+                   lock_detect_params.lp, lock_detect_params.lo);
+
+  /* TODO: Reconfigure alias detection between stages */
+  u8 alias_detect_ms = l->coherent_ms;
+  alias_detect_init(&data->alias_detect,
+                    L2C_ALIAS_DETECT_INTERVAL_MS / alias_detect_ms,
+                    (alias_detect_ms - 1) * 1e-3);
+
+  /* Let's remember SV ID to avoid double tracking of L2C from it. */
+  data->sat = channel_info->sid.sat;
```
jacobmcnamee: Is this used anywhere? I don't think it is needed.

src/track/track_gps_l2cm.c
```diff
@@ -0,0 +1,614 @@
...
+    return;
+  }
+
+  common_data->update_count += data->int_ms;
+
+  /* L2C bit sync is known once we start tracking it since
+     the L2C ranging code length matches the bit length (20ms).
+     This is the end of 20ms integration period and the edge
+     of a data bit. */
+  tracker_bit_sync_set(channel_info->context, 0);
```
jacobmcnamee: Is it reasonable to call this once when tracking is first started?

src/track/track_gps_l2cm.c
```diff
@@ -0,0 +1,614 @@
...
+    .code_phase         = code_phase,
+    .chips_to_correlate = L2CM_TRACK_SHORT_CYCLE_INTERVAL_CHIPS,
+    .cn0_init           = cn0_init,
+    .elevation          = TRACKING_ELEVATION_UNKNOWN
+  };
+
+  if (tracking_startup_request(&startup_params)) {
+    log_info_sid(sid, "L2 CM handover done");
+  } else {
+    log_error_sid(sid, "Failed to start L2C tracking");
```
jacobmcnamee: This should be an info or warning.

src/track/track_gps_l2cm.c
```diff
@@ -0,0 +1,614 @@
...
+
+#include "board.h"
+#include "nap/nap_common.h"
+#include "nap/track_channel.h"
+#include "track.h"
+#include "main.h"
+
+#include "track.h"
+
+#define FEATURE_L2C  /* skip weak attributes for L2C API implementation */
```
jacobmcnamee: Maybe make the define more specific so you don't have to worry about the undef: `TRACK_GPS_L2CM_INTERNAL` ?

src/track/track_gps_l2cm.c
```diff
@@ -0,0 +1,614 @@
...
+#include <libswiftnav/track.h>
+
+#include <string.h>
+#include <assert.h>
+#include "math.h"
+
+#include "settings.h"
+#include "signal.h"
+
+#define NUM_GPS_L2CM_TRACKERS   32
```
jacobmcnamee: move define to board/v3/platform_signal.h

src/board/v3/nap/track_channel.c
adel-mamin: I have moved sid_to_init_g1() to prn.c in libswiftnav.

src/track/track_gps_l2cm.c
adel-mamin: Agree. I will leave the code as is then.

src/track/track_gps_l2cm.c
adel-mamin: Should be fixed now.

src/track/track_gps_l2cm.c
adel-mamin: My bad. Should not be here at all. It is a leftover from the older approach, where I checked if the SV L2CM signal is already in track. Anyways, now it is removed.

src/track/track_gps_l2cm.c
adel-mamin: No, it is not reasonable. You are right. I moved it to the tracker_gps_l2cm_init().

src/track/track_gps_l2cm.c
adel-mamin: Should be fixed now.

src/track/track_gps_l2cm.c
adel-mamin: Agree. I changed it to FEATURE_TRACK_GPS_L2CM and removed the 'undef'.

src/track/track_gps_l2cm.c
adel-mamin: Done.

src/track/track_gps_l2cm.c
adel-mamin: Should be fixed now.

src/board/v3/nap/nap_hw.h
```diff
@@ -107,8 +107,31 @@ typedef struct {
...
+enum {
+  NAP_RF_FRONTEND_CHANNEL_1 = 0, /* GPS L1 */
+  NAP_RF_FRONTEND_CHANNEL_2,     /* Glonass L1 */
+  NAP_RF_FRONTEND_CHANNEL_3,     /* Glonass L2 */
+  NAP_RF_FRONTEND_CHANNEL_4      /* GPS L2 */
+};
+
+/* The values of NAP_TRKx_CONTROL::CODE field */
+enum {
+  NAP_CONSTELLATION_BAND_1 = 0, /* GPS L1C/A, SBAS L1CA */
```
jacobmcnamee: `NAP_CODE_GPS_L1CA_SBAS_L1CA`
`NAP_CODE_GPS_L2CM` ?

src/board/v3/nap/track_channel.c
```diff
@@ -50,45 +51,111 @@ static struct {
...
+  track_count += (SAMPLE_FREQ / code_to_chip_rate(sid.code)) *
+                 (code_to_chip_count(sid.code) - cp) *
+                 (1.0 + carrier_freq / code_to_carr_freq(sid.code));
 
   u8 prn = sid.sat - 1;
-  NAP->TRK_CH[channel].CONTROL = prn << NAP_TRK_CONTROL_SAT_Pos;
+  /* PRN code */
+  control = (prn << NAP_TRK_CONTROL_SAT_Pos) & NAP_TRK_CONTROL_SAT_Msk;
+  /* RF frontend channel */
+  control |= (sid_to_rf_frontend_channel(sid) << NAP_TRK_CONTROL_RF_FE_Pos) & \
```
jacobmcnamee: remove backslashes

src/track.c
```diff
@@ -992,18 +998,20 @@ static void event(tracker_channel_t *tracker_channel, event_t event)
...
   memset(common_data, 0, sizeof(tracker_common_data_t));
 
   common_data->TOW_ms = TOW_INVALID;
 
   /* Calculate code phase rate with carrier aiding. */
-  common_data->code_phase_rate = (1 + carrier_freq/GPS_L1_HZ) *
-                                    GPS_CA_CHIPPING_RATE;
+  common_data->code_phase_rate = (1 + carrier_freq / code_to_carr_freq(code)) *
+                                 GPS_CA_CHIPPING_RATE;
   common_data->carrier_freq = carrier_freq;
```
jacobmcnamee: Ugh we really need to go through and rename "carrier frequency" to "carrier doppler" or something 100% less misleading at some point.

src/track/track_gps_l1ca.c
```diff
@@ -337,11 +344,20 @@ static void tracker_gps_l1ca_update(const tracker_channel_info_t *channel_info,
 
     /* Indicate that a mode change has occurred. */
     common_data->mode_change_count = common_data->update_count;
+
+    do_l1ca_to_l2cm_handover(common_data->sample_count,
```
jacobmcnamee: What happens if the L2CM tracking channel fails to start or is dropped? It should probably attempt handover periodically, though I'm not sure the best way to accomplish that.

src/track.c
adel-mamin: I agree. That would be great. It confused me already many times. And in this particular place it is as confusing as it could be.

src/track/track_gps_l2cm.c
```diff
@@ -0,0 +1,617 @@
...
+       the previous full 20 ms cycle. */
+    u32 sample_count;            /**< Total num samples channel has tracked for. */
+    double code_phase_early;     /**< Early code phase. */
+    double carrier_phase;        /**< Carrier phase in NAP register units. */
+
+    tracker_correlations_read(channel_info->context, data->cs,
+                              &sample_count,
+                              &code_phase_early,
+                              &carrier_phase);
+    (void) sample_count;
```
jacobmcnamee: Shouldn't need the `(void) var;` stuff

src/track/track_gps_l1ca.c
adel-mamin: The issue of L2CM re-acquisition should be addressed by another PR https://github.com/swift-nav/piksi_firmware/pull/681. In that PR it indeed tries to re-acquire L2CM periodically unless it is already in track.

src/track/track_gps_l2cm.h
```diff
@@ -0,0 +1,34 @@
...
+#ifdef FEATURE_TRACK_GPS_L2CM
+# define L2C_WEAK
+#else
+# define L2C_WEAK __attribute__ ((weak, alias ("l2c_not_implemented")))
+#endif  /* FEATURE_TRACK_GPS_L2CM */
+
+int l2c_not_implemented() __attribute__ ((weak));
+inline int l2c_not_implemented() { return -1; }
+
+void track_gps_l2cm_register(void) L2C_WEAK;
```
jacobmcnamee: Shouldn't need `L2C_WEAK` for this function

src/track/track_gps_l2cm.h
```diff
@@ -0,0 +1,34 @@
...
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+#ifndef SWIFTNAV_TRACK_GPS_L2CM_H
+#define SWIFTNAV_TRACK_GPS_L2CM_H
+
+#include <libswiftnav/common.h>
+
+#ifdef FEATURE_TRACK_GPS_L2CM
```
jacobmcnamee: I still like `TRACK_GPS_L2CM_INTERNAL` or similar so that it's clear that users should never define this value.

src/manage.c
```diff
@@ -298,6 +299,12 @@ static acq_status_t * choose_acq_sat(void)
   gps_time_t t = get_current_time();
 
   for (u32 i=0; i<PLATFORM_SIGNAL_COUNT; i++) {
+    if (CODE_GPS_L2CM == acq_status[i].sid.code) {
```
jacobmcnamee: Can you put this in a function which returns true if an SID should be excluded from direct acquisition?

jacobmcnamee: For consistency I'd also make the same check in the second loop later in this function

src/track/track_gps_l2cm.c
```diff
@@ -0,0 +1,617 @@
...
+    return;
+  }
+
+  /* Read early ([0]), prompt ([1]) and late ([2]) correlations. */
+  if (data->short_cycle) {
+    /* The throw away data. Not needed for the short cycle.
+       And we also do not want to clobber common_data content by
+       by writing these data to it as it contains valid data for
+       the previous full 20 ms cycle. */
+    u32 sample_count;            /**< Total num samples channel has tracked for. */
```
jacobmcnamee: probably don't want doxygen comments for these

jacobmcnamee:
I haven't gone over the overall logic in `tracker_gps_l2cm_update()` super thoroughly, but I imagine a lot of it is copied from GPS L1CA. Is there anything in particular you want checked? Otherwise it's looking great.

adel-mamin:
You are right. The general structure is copied from GPS L1C/A. I have no special requests. You've already provided many valuable comments. I will fix those and notify you once I am done.

src/board/v3/nap/nap_hw.h
adel-mamin: Should be fixed now.

src/board/v3/nap/track_channel.c
adel-mamin: Should be fixed now.

src/track/track_gps_l2cm.c
adel-mamin: Should be fixed now.

src/track/track_gps_l2cm.h
adel-mamin: Should be fixed now.

src/track/track_gps_l2cm.h
adel-mamin: Should be fixed now.

src/manage.c
adel-mamin: I added the new API function code_requires_direct_acq() into libswiftnav. Please check https://github.com/swift-nav/libswiftnav-private/pull/6: 
https://github.com/adel-mamin/libswiftnav-private/blob/9ce04464eb17c7a76e7882b83ac4b5e20a7bf52d/src/signal.c#L280

src/track/track_gps_l2cm.c
adel-mamin: Should be fixed now.

src/track/track_gps_l2cm.c
```diff
@@ -0,0 +1,614 @@
...
+    }
+    if (data->startup) {
+      data->startup--;
+    }
+    return;
+  }
+
+  /* Read early ([0]), prompt ([1]) and late ([2]) correlations. */
+  if (data->short_cycle) {
+    /* The throw away data. Not needed for the short cycle.
```
jacobmcnamee: I think it's fine to update these values in the `common_data` struct; that's what L1CA does. Either way we need to make sure it's all or nothing, at least for the critical parameters (`sample_count`, `code_phase`, `carrier_phase`, `time_of_week_ms`)

adel-mamin: The short cycle on L2C is set to ~300 us at the moment (L2CM_TRACK_SHORT_CYCLE_INTERVAL_CHIPS). The long cycle is ~19700 us. The TOW is only updated after the long cycle is over. So I do not call  tracker_tow_update() after the short cycle, because tracker_tow_update expects full ms updates to TOW. Since TOW is not updated on the short cycle, I do not update common_data struct either to avoid the inconsistency.

rge-exafore:
Moved to private and closed.

PR #674, opened 2016-05-12
Compute measurement time delta in tracking_channel_measurement_get()

jacobmcnamee:
Compute the integer difference between `nav_tc` and tracking channel timing count inside `tracking_channel_measurement_get()` instead of converting both to doubles and subtracting in libswiftnav. Better numerically and facilitates computing 64bit timing count for the tracking channel for #651.

### DIFF
--- a/src/solution.c
+++ b/src/solution.c
@@ -411,12 +411,13 @@ static void solution_thread(void *arg)
       solution_simulation();
     }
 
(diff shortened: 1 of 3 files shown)


jacobmcnamee:
Requires https://github.com/swift-nav/libswiftnav/pull/347

PR #675, opened 2016-05-13
NMEA update

switanis:
This is the first round of NMEA module update. Changes:
- small code cleanup
- comments updated, message examples removed
- lat/lon computation is changed to avoid case where number of minutes is rounded to 60 in the final snprintf()
- reported year in ZDA message is corrected
- order of outputting messages is changed to output more important messages first 

Cc: @denniszollo @swift-nav/firmware

### DIFF
--- a/src/nmea.c
+++ b/src/nmea.c
@@ -167,22 +167,24 @@ void nmea_gpgga(const double pos_llh[3], const gps_time_t *gps_t, u8 n_used,
 
   unix_t = gps2time(gps_t);
   gmtime_r(&unix_t, &t);
+ 
+  double frac_s  = fmod(gps_t->tow, 1.0);
+  
+  double lat     = fabs( round( R2D * pos_llh[0] * 1e8 ) / 1e8 );
+  double lon     = fabs( round( R2D * pos_llh[1] * 1e8 ) / 1e8 );
 
-  double frac_s = fmod(gps_t->tow, 1.0);
+  char   lat_dir = pos_llh[0] < 0.0 ? 'S' : 'N';
+  u16    lat_deg = (u16)lat;
+  double lat_min = (lat - (double)lat_deg) * 60.0;
 
-  s16 lat_deg = R2D * (pos_llh[0]);
-  double lat_min = MINUTES(pos_llh[0]);
-  s16 lon_deg = R2D * (pos_llh[1]);
-  double lon_min = MINUTES(pos_llh[1]);
-  lat_deg = abs(lat_deg);
-  lon_deg = abs(lon_deg);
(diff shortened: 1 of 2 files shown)


src/nmea.c
```diff
@@ -54,7 +55,8 @@ static struct nmea_dispatcher *nmea_dispatchers_head;
  * max_len = max possible length of the body of the message
  * (not including suffix)
  */
-#define NMEA_SENTENCE_START(max_len) char sentence_buf[max_len + NMEA_SUFFIX_LEN]; \
+#define NMEA_SENTENCE_START(max_len) \
```
denniszollo: should we put this macro in a do{} while loop?

switanis: I don't think there is a need for that.

jacobmcnamee: It wouldn't work, anyway. The allocated variables need to be visible to the other macros.

src/nmea.c
```diff
@@ -167,44 +174,44 @@ void nmea_gpgga(const double pos_llh[3], const gps_time_t *gps_t, u8 n_used,
...
-  lon_deg = abs(lon_deg);
-  char lat_dir = pos_llh[0] < 0 ? 'S' : 'N';
-  char lon_dir = pos_llh[1] < 0 ? 'W' : 'E';
-
-  NMEA_SENTENCE_START(120);
-  /* Note, geoid separation is set to 0 because no geoid model exists*/
+ 
+  double frac_s  = fmod(gps_t->tow, 1.0);
+  
+  double lat     = fabs( round( R2D * pos_llh[0] * 1e8 ) / 1e8 );
```
jacobmcnamee: Make this a macro? Seems to be repeated in many places.

jacobmcnamee:
Please remove spaces inside parens, e.g. 
```
if ( a > b ) -> if (a > b)
```

src/nmea.c
switanis: In the next round. In fact all NMEA values should be computed once at the beginning of NMEA creation block and then used for all messages as needed. Code will be simpler, data will be consistent by design and run faster.

switanis:
Spaces improve visibility and readability and as such make code less error prone.

denniszollo:
LGTM after following style guide

switanis:
HITL results:
http://sbp-log-analysis.swiftnav.com/#/d/0/q/x/firmware/y/metric/f/metric/p/all/f/firmware/sv/piksi_firmware_v0.21.hex%2Cpiksi_firmware_v0.21-205-g253dada.hex/st/1?_k=5olsfm

denniszollo:
Plan to merge at 4:05.  Please get in any feedback if you see issues!

@swift-nav/firmware

jacobmcnamee:
lgtm

PR #683, opened 2016-05-18
v3: NAPv3 refactoring

wltr:
Modify the register map to be compatible with NAPv3 refactoring. The
frontend DSP chain was moved to its own IP block with a separate
register map.

/cc @jacobmcnamee @gsmcmullin


wltr:
I'm closing this as I'm combining several PRs into a new rebased one.

PR #685, opened 2016-05-19
Check for high RTK DOPs

ljbade:
This PR checks for RTK solutions with PDOPs > 20 and prevents a solution being sent.

I hope this will fix occasional float solution outliers that have been observed recently up to 200km in error.

The current firmware only checks the DOPs of the SPP solution, but due to cycle slips or missing base observations the RTK solution might use less satellites than SPP and try to output solutions with very high DOPs.

This PR is based on https://github.com/swift-nav/piksi_firmware/pull/651 so that must be merged first.

Requires libswiftnav with https://github.com/swift-nav/libswiftnav/pull/351

/cc @swift-nav/firmware @swift-nav/estimation @denniszollo @anth-cole

### DIFF
--- a/src/solution.c
+++ b/src/solution.c
@@ -78,7 +78,7 @@ static u16 lock_counters[PLATFORM_SIGNAL_COUNT];
 bool disable_raim = false;
 bool send_heading = false;
 
-void solution_send_sbp(gnss_solution *soln, dops_t *dops)
+void solution_send_sbp(gnss_solution *soln, dops_t *dops, bool clock_jump)
 {
   if (soln) {
     /* Send GPS_TIME message first. */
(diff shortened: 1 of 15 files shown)


src/solution.c
```diff
@@ -254,7 +262,30 @@ static void output_baseline(u8 num_sdiffs, const sdiff_t *sdiffs,
...
+    /* The DOPs are always valid even if calc_PVT fails. */
+    calc_PVT(intersection_size, intersection_nm, TRUE, &soln, &dops);
+    log_debug("RTK DOPs: H %.2f V %.2f P %.2f T %.2f G %.2f",
+              dops.hdop, dops.vdop, dops.pdop, dops.tdop, dops.gdop);
+
+    /* Ensure PDOP is not too high. */
+    if (dops.pdop <= 20.0f) {
+      solution_send_baseline(t, num_used, b, position_solution.pos_ecef, flags, dops.hdop, diff_time, base_id);
+    } else {
+      log_error("output_baseline: RTK PDOP %.2f > 20", dops.pdop);
```
denniszollo: should this really be a log_error? I think  a log_warn will suffice.

ljbade: Sure, what error level is the SPP errors logged at? Since that also triggers on high DOP.

ljbade:
Making a note from some testing yesterday, need to change SPP cutoff back to 50.

ljbade:
@denniszollo I think it might be valuable to also output the RTK DOP in `MSG_DOPS`... perhaps add a new message `MSG_RTK_DOPS`

PR #686, opened 2016-05-19
WIP: Add Iono correction

dt-exafore:
This adds iono correction.
Note, iono parameter decoder should be merged first https://github.com/swift-nav/libswiftnav-private/pull/8

### DIFF
--- a/src/base_obs.c
+++ b/src/base_obs.c
@@ -35,6 +35,7 @@
 #include "base_obs.h"
 #include "ephemeris.h"
 #include "signal.h"
+#include "iono.h"
 
 extern bool disable_raim;
 
@@ -145,6 +146,21 @@ static void update_obss(obss_t *new_obss)
   if (base_obss.n >= 4) {
     gnss_solution soln;
     dops_t dops;
+
+    /* check if we have fix, if yes, calculate iono and tropo correction */
+    if(base_obss.has_pos) {
+      double llh[3];
+      wgsecef2llh(base_obss.pos_ecef, llh);
+      log_debug("Base: IONO/TROPO correction");
+      ionosphere_t i_params;
+      ionosphere_t *p_i_params = &i_params;
+      /* get iono parameters if available */
+      if(!gps_iono_params_read(p_i_params)) {
+        p_i_params = NULL;
+      }
+      calc_iono_tropo(base_obss.n, base_obss.nm, base_obss.pos_ecef, llh,
+                      p_i_params);
+    }
 
     /* Calculate a position solution. */
     /* disable_raim controlled by external setting (see solution.c). */
--- a/src/decode/decode_gps_l1ca.c
+++ b/src/decode/decode_gps_l1ca.c
@@ -23,6 +23,7 @@
 #include "sbp_utils.h"
 #include "signal.h"
 #include "l2c_capb.h"
+#include "iono.h"
 
 typedef struct {
   nav_msg_t nav_msg;
@@ -113,6 +114,12 @@ static void decoder_gps_l1ca_process(const decoder_channel_info_t *channel_info,
     /* Store new L2C value  */
     log_debug("L2C capabilities received: 0x%x", dd.gps_l2c_sv_capability);
     gps_l2cm_l2c_cap_store(dd.gps_l2c_sv_capability);
+  }
+
+  if (dd.iono_corr_upd_flag) {
+    /* store new iono parameters */
+    log_debug("Iono parameters received");
+    gps_iono_params_store(&dd.iono);
   }
 
   if(dd.ephemeris_upd_flag) {
--- a/src/iono.c
+++ b/src/iono.c
@@ -1,0 +1,53 @@
+/*
+ * Copyright (C) 2016 Swift Navigation Inc.
+ * Contact: Dmitry Tatarinov <dmitry.tatarinov@exafore.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#include <assert.h>
+#include <string.h>
+#include <ch.h>
+
+#include "iono.h"
+#include "signal.h"
+
+MUTEX_DECL(iono_mutex);
+
+static ionosphere_t iono_params;
+static bool iono_params_decoded_flag = false;
+
+/** Stores ionospheric parameters
+ * \param params pointer to ionospheric parameters to be stored
+ */
+void gps_iono_params_store(ionosphere_t *params)
+{
+  assert(params != NULL);
+  chMtxLock(&iono_mutex);
+  memcpy(&iono_params, params, sizeof(ionosphere_t));
+  iono_params_decoded_flag = true;
+  chMtxUnlock(&iono_mutex);
+}
+
+/** Reads ionospheric parameters
+ * \param params pointer to ionospheric
+ * \return 1 if iono parameters available otherwise 0
+ */
+u8 gps_iono_params_read(ionosphere_t *params)
+{
+  assert(params != NULL);
+  chMtxLock(&iono_mutex);
+  if (iono_params_decoded_flag) {
+    memcpy(params, &iono_params, sizeof(ionosphere_t));
+    chMtxUnlock(&iono_mutex);
+    return 1;
+  } else {
+    chMtxUnlock(&iono_mutex);
+    return 0;
+  }
+}
--- a/src/solution.c
+++ b/src/solution.c
@@ -43,6 +43,7 @@
 #include "signal.h"
 #include "system_monitor.h"
 #include "main.h"
+#include "sid_set.h"
 
 /* Maximum CPU time the solution thread is allowed to use. */
 #define SOLN_THD_CPU_MAX (0.60f)
@@ -462,7 +463,12 @@ static void solution_thread(void *arg)
       tracking_channel_unlock(i);
     }
 
-    if (n_ready < 4) {
+    gnss_sid_set_t codes_in_track;
+    sid_set_init(&codes_in_track);
+    for (u8 i=0; i<n_ready; i++)
+      sid_set_add(&codes_in_track, meas[i].sid);
+
+    if (sid_set_get_sat_count(&codes_in_track) < 4) {
       /* Not enough sats, keep on looping. */
       continue;
     }
@@ -517,7 +523,13 @@ static void solution_thread(void *arg)
     n_ready_old = n_ready;
     rec_tc_old = rec_tc;
 
-    if (n_ready_tdcp < 4) {
+    gnss_sid_set_t codes_tdcp;
+    sid_set_init(&codes_tdcp);
+    for (u8 i=0; i<n_ready_tdcp; i++) {
+      sid_set_add(&codes_tdcp, nav_meas_tdcp[i].sid);
+    }
+
+    if (sid_set_get_sat_count(&codes_tdcp) < 4) {
       /* Not enough sats to compute PVT */
       continue;
     }
--- a/src/solution.c
+++ b/src/solution.c
@@ -654,7 +654,8 @@ static void solution_thread(void *arg)
           memcpy(nm, &nav_meas_tdcp[i], sizeof(*nm));
         }
 
-        nm->raw_pseudorange += t_err * nm->raw_doppler * GPS_L1_LAMBDA;
+        nm->raw_pseudorange += t_err * nm->raw_doppler *
+                               code_to_lambda(nm->sid.code);
         nm->raw_carrier_phase += t_err * nm->raw_doppler;
 
         nm->tot = new_obs_time;
--- a/src/solution.c
+++ b/src/solution.c
@@ -9,7 +9,6 @@
  * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
  * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
  */
-
 #include <stdio.h>
 #include <string.h>
 
@@ -23,6 +22,7 @@
 #include <libswiftnav/dgnss_management.h>
 #include <libswiftnav/baseline.h>
 #include <libswiftnav/linear_algebra.h>
+#include <libswiftnav/troposphere.h>
 
 #define memory_pool_t MemoryPool
 #include <ch.h>
@@ -43,6 +43,7 @@
 #include "signal.h"
 #include "system_monitor.h"
 #include "main.h"
+#include "iono.h"
 #include "sid_set.h"
 
 /* Maximum CPU time the solution thread is allowed to use. */
@@ -433,6 +434,9 @@ static void sol_thd_sleep(systime_t *deadline, systime_t interval)
 static THD_WORKING_AREA(wa_solution_thread, 8000);
 static void solution_thread(void *arg)
 {
+  /* The flag is true when we have a fix */
+  bool soln_flag = false;
+
   (void)arg;
   chRegSetThreadName("solution");
 
@@ -534,6 +538,20 @@ static void solution_thread(void *arg)
       continue;
     }
 
+    /* check if we have a solution, if yes calc iono and tropo correction */
+    if (soln_flag) {
+      ionosphere_t i_params;
+      ionosphere_t *p_i_params = &i_params;
+      /* get iono parameters if available */
+      if(!gps_iono_params_read(p_i_params)) {
+        p_i_params = NULL;
+      }
+      calc_iono_tropo(n_ready_tdcp, nav_meas_tdcp,
+                      position_solution.pos_ecef,
+                      position_solution.pos_llh,
+                      p_i_params);
+    }
+
     dops_t dops;
     /* Calculate the SPP position
      * disable_raim controlled by external setting. Defaults to false. */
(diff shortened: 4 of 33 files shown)


ljbade:
Not sure how far away the nav DB stuff is.

But in the meantime a static variable in ionosphere.c with appropriate mutex would be good enough to allow testing the ionosphere correction.

dt-exafore:
@ljbade yes, since NDB not merged yet I use store/read API in new file iono.c, by analogy to l2c capability.
My concern about at what stage I should add iono correction?
The idea is to add the correction to pseudorange into calc_navigation_measurement. Once we've got decoded parameters and had a fix we can calculate ionosphere delay using calc_ionosphere() and subtract it from  pseudorange. What do you think?

ljbade:
@dt-exafore Yeah there is a non raw `pseudorange` and `carrierphase` field of `navigation_measurement` (however before #651 the current carrierphase field is actually the "raw" value).

The non-raw field is for use with PVT calculations and includes other corrections such as the satellite clock error. The raw field is for RTK and should be left as is.

I can see two possible places to apply the correction either `calc_navigation_measurement` or `calc_PVT` (in pvt.c).

I think the best choice will be wherever it is easiest to pass in the prior PVT position solution which will be needed as input into the tropo and iono model.

dt-exafore:
@ljbade Could you please check the output of calc_ionosphere function I have using latest changes?
I'm not sure they are correct.
INFO Iono correction for channel 0: 8658123314.92751121520996093750 m
INFO Iono correction for channel 1: 8913394006.64842605590820312500 m
INFO Iono correction for channel 2: 8111396194.80219936370849609375 m
INFO Iono correction for channel 3: 5395831060.94595336914062500000 m

ljbade:
@dt-exafore that seems way too high, what are the alpha/beta parameter values?

src/iono.c
```diff
@@ -0,0 +1,49 @@
...
+
+/** Stores ionospheric parameters
+ * \param ch_id Tracking channel ID
+ * \param params pointer to ionospheric parameters to be stored
+ */
+void gps_iono_params_store(u8 ch_id, ionosphere_t *params)
+{
+  assert(ch_id < NUM_TRACKER_CHANNELS);
+  assert(params != NULL);
+  memcpy(&iono_params[ch_id], params, sizeof(ionosphere_t));
```
ljbade: I think this will need a mutex to prevent store and read functions being called at same time.

src/solution.c
```diff
@@ -468,12 +472,37 @@ static void solution_thread(void *arg)
       continue;
     }
 
+    /* check if we have a solution, if yes calc iono correction */
+    if (soln_flag) {
+      for (u8 i = 0; i < n_ready_tdcp; i++) {
+        ionosphere_t iono_params;
+        double az, el, iono_correction;
+        /* read iono parameters for the tracking channel if available */
+        if (gps_iono_params_read(i, &iono_params)) {
```
ljbade: `i` is not necessarily the channel ID

src/iono.c
```diff
@@ -0,0 +1,49 @@
...
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#include <assert.h>
+#include <string.h>
+
+#include "iono.h"
+#include "signal.h"
+
+static ionosphere_t iono_params[NUM_TRACKER_CHANNELS];
```
ljbade: Do we need one set of params per channel? We should only need one value for all GPS sats (and GLONASS sats too).

src/iono.c
```diff
@@ -0,0 +1,49 @@
...
+ * \param ch_id Tracking channel ID
+ * \param params pointer to ionospheric
+ * \return 1 if iono parameters available otherwise 0
+ */
+u8 gps_iono_params_read(u8 ch_id, ionosphere_t *params)
+{
+  assert(ch_id < NUM_TRACKER_CHANNELS);
+  assert(params != NULL);
+  if (iono_params_decoded_flag & (1 << ch_id)) {
+    params = &iono_params[ch_id];
```
ljbade: this will not result in the values being copied out of the function
you need to use either `memcpy` or pass a pointer to a pointer...

src/solution.c
```diff
@@ -468,12 +472,37 @@ static void solution_thread(void *arg)
...
+          wgsecef2azel(nav_meas[i].sat_pos, position_solution.pos_ecef,
```
ljbade: Also here you use `nav_meas` but `calc_ionosphere` uses `nav_meas_tdcp[i].tot`

src/solution.c
```diff
@@ -468,12 +471,35 @@ static void solution_thread(void *arg)
...
     }
 
+    /* check if we have a solution, if yes calc iono correction */
+    if (soln_flag) {
+      for (u8 i = 0; i < n_ready_tdcp; i++) {
+        ionosphere_t iono_params;
+        double az, el, iono_correction;
+        /* read iono parameters for the tracking channel if available */
+        if (gps_iono_params_read(&iono_params)) {
+          wgsecef2azel(nav_meas[i].sat_pos, position_solution.pos_ecef,
```
ljbade: This should use `nav_meas_tdcp`

src/iono.c
```diff
@@ -0,0 +1,54 @@
...
+/** Reads ionospheric parameters
+ * \param params pointer to ionospheric
+ * \return 1 if iono parameters available otherwise 0
+ */
+u8 gps_iono_params_read(ionosphere_t *params)
+{
+  assert(params != NULL);
+  chMtxLock(&iono_mutex);
+  if (iono_params_decoded_flag) {
+    memset(params, 0, sizeof(ionosphere_t));
```
ljbade: Why is memset needed before memcpy?

src/solution.c
```diff
@@ -468,12 +471,35 @@ static void solution_thread(void *arg)
...
+                       &az, &el);
+          /* calculate iono correction */
+          iono_correction = calc_ionosphere(&nav_meas_tdcp[i].tot,
+                                            position_solution.pos_llh[0],
+                                            position_solution.pos_llh[1],
+                                            az,el,
+                                            &iono_params);
+          /* correct pseudorange */
+          nav_meas_tdcp[i].pseudorange -= iono_correction;
```
ljbade: After https://github.com/swift-nav/piksi_firmware/pull/651 lands (later today), you should also add a opposite sign correction for the carrier phase.

src/iono.c
dt-exafore: Well, since params it's a pointer to stack memory which is allocated during run time, I thought it's good to clear it first, but probably for this particular case it's not needed because I copy whole structure. I'll remove that.

src/solution.c
dt-exafore: Now I use one iono parameters for all channels.

src/solution.c
dt-exafore: @ljbade what is the criteria to change the sign?

dt-exafore:
Latest tests results
INFO Iono correction for channel 0:    4.6536859986637741926074341
INFO Iono correction for channel 1:    3.1751344664971745679338255
INFO Iono correction for channel 2:    4.3321395864821621302098720
INFO Iono correction for channel 3:    4.7696433655641463644769829

dt-exafore:
@nsirola, please also review.

src/solution.c
```diff
@@ -491,8 +491,7 @@ static void solution_thread(void *arg)
           /* correct carrier phase */
           cp_correction = iono_correction * (GPS_L1_HZ + nav_meas_tdcp[i].doppler)
                           / GPS_C;
-          log_info("CP corr %u: %30.20f", i, cp_correction);
-          nav_meas_tdcp[i].carrier_phase += cp_correction;
+          nav_meas_tdcp[i].carrier_phase -= cp_correction;
```
ljbade: I think it would also be wise to add a comment that the sign here is opposite to normal due to Piksi's unusual sign convention on carrier phase.

In case this confuses anyone.

nsirola:
Looks good to me.

dt-exafore:
Measurements during tests:
INFO Channel 0: IONO         5.71110030471377339012, TROPO        11.68359588241024660249
INFO Channel 1: IONO         5.84155407125674130953, TROPO         5.00802388949610044477
INFO Channel 2: IONO         5.48495403509538625997, TROPO         5.92363907484279472726
INFO Channel 3: IONO         6.58029111723088178110, TROPO         6.05548545450361253017

ljbade:
When L2C tracking is completed, the iono corrections will need to be adjusted for the different carrier frequencies.

There is a formula to convert from the L1 Klobuchar correction, to any other frequency:
image
(from http://www.navipedia.net/index.php/Klobuchar_Ionospheric_Model)

Also the carrier phase adjustments will need to use the carrier frequency of each signal.

src/solution.c
```diff
@@ -468,12 +472,49 @@ static void solution_thread(void *arg)
...
+          /* calculate iono correction */
+          iono_correction = calc_ionosphere(&nav_meas_tdcp[i].tot,
+                                            position_solution.pos_llh[0],
+                                            position_solution.pos_llh[1],
+                                            az,el,
+                                            &iono_params);
+          /* calculate tropo correction */
+          tropo_correction = calc_troposphere(&nav_meas_tdcp[i].tot,
+                                              position_solution.pos_llh[0],
+                                              position_solution.pos_llh[2],
```
ljbade: I think there should also be a note that we are currently using the ellipsoidal height as an approximation to the orthometric (or geoid) height due to lack of a geoid model.

src/solution.c
```diff
@@ -468,12 +472,49 @@ static void solution_thread(void *arg)
...
+                                              el);
+          /* correct pseudorange */
+          nav_meas_tdcp[i].pseudorange -= tropo_correction + iono_correction;
+          /* correct carrier phase based on iono correction */
+          cp_correction = iono_correction * (GPS_L1_HZ + nav_meas_tdcp[i].doppler)
```
ljbade: I think it would also be wise to add a comment that the sign here is opposite to normal due to Piksi's unusual sign convention on carrier phase.

In case this confuses anyone

ljbade:
Having some compile problems. The PR needs to be rebased on master. Also libswiftnav submodule needs to updated.

src/solution.c
```diff
@@ -468,12 +472,49 @@ static void solution_thread(void *arg)
       continue;
     }
 
+    /* check if we have a solution, if yes calc iono correction */
+    if (soln_flag) {
+      for (u8 i = 0; i < n_ready_tdcp; i++) {
+        ionosphere_t iono_params;
+        double az, el, iono_correction, tropo_correction, cp_correction;
+        /* read iono parameters for the tracking channel if available */
+        if (gps_iono_params_read(&iono_params)) {
```
ljbade: This check should only wrap `calc_ionosphere` as `calc_troposphere` does not need the iono params.

src/solution.c
```diff
@@ -468,12 +472,49 @@ static void solution_thread(void *arg)
       continue;
     }
 
+    /* check if we have a solution, if yes calc iono correction */
+    if (soln_flag) {
+      for (u8 i = 0; i < n_ready_tdcp; i++) {
```
benjamin0: could we pull these corrections into a separate pre-processing function?

ljbade: Good idea... then we could also call the function for base_obs.c too

src/solution.c
```diff
@@ -487,6 +539,10 @@ static void solution_thread(void *arg)
       continue;
     }
 
+    /* check if we have a solution, if yes calc iono correction */
+    if (soln_flag)
+      calc_iono_tropo(n_ready_tdcp, nav_meas_tdcp);
```
ljbade: can you use curly braces for the if statement

src/solution.c
```diff
@@ -396,6 +401,53 @@ static void sol_thd_sleep(systime_t *deadline, systime_t interval)
   chSysUnlock();
 }
 
+/** Calculates Iono and Tropo correction
+ * \param n_ready_tdcp The number of measurements written to `m_tdcp`
+ * \param nav_meas_tdcp Pointer to measurements array */
+void calc_iono_tropo(u8 n_ready_tdcp,
```
ljbade: Can you also call this in base_obs.c so our base positions are also improved:
Just before here https://github.com/swift-nav/piksi_firmware/blob/master/src/base_obs.c#L189

benjamin0: should we hold off on these corrections for RTK for now? we might want to formalize our pre processing steps to make sure we're transmitting/logging raw pseudoranges but using corrected ones in the filter. with a short baseline they won't have much of an effect anyway

ljbade: @benjamin0 In our navigation_message structures we very clearly separate pseudrorange / raw_pseudorange and carrier_phase / raw_carrier_phase.

Anything to do with RTK should be using raw_* values, only SPP uses the non-raw values.

src/solution.c
```diff
@@ -78,6 +80,9 @@ static u16 lock_counters[PLATFORM_SIGNAL_COUNT];
 bool disable_raim = false;
 bool send_heading = false;
 
+/* The flag is true when we have a fix */
+static bool soln_flag = false;
```
pmiettinen: This variable seems to be used only inside solution_thread loop. Wouldn't it be better to declare this variable inside the thread function instead of file scope global?

ljbade: Yeah that would make sense

src/base_obs.c
```diff
@@ -146,6 +146,14 @@ static void update_obss(obss_t *new_obss)
     gnss_solution soln;
     dops_t dops;
 
+    /* check if we have fix, if yes, calculate iono and tropo correction */
+    if(base_obss.has_pos) {
+      double llh[3];
+      wgsecef2llh(base_obss.pos_ecef, llh);
+      log_info("Base: IONO/TROPO correction");
```
ljbade: Will need to remove in final PR (or make debug)

src/decode/decode_gps_l1ca.c
```diff
@@ -115,6 +116,12 @@ static void decoder_gps_l1ca_process(const decoder_channel_info_t *channel_info,
     gps_l2cm_l2c_cap_store(dd.gps_l2c_sv_capability);
   }
 
+  if (dd.iono_corr_upd_flag) {
+    /* store new iono parameters */
+    log_info("Iono parameters received");
```
ljbade: Will need to remove in final PR (or make debug)

src/solution.c
```diff
@@ -9,7 +9,7 @@
  * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
  * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
  */
-
+#define DEBUG 1
```
ljbade: Will need to remove in final PR

src/solution.c
```diff
@@ -429,6 +434,56 @@ static void sol_thd_sleep(systime_t *deadline, systime_t interval)
   chSysUnlock();
 }
 
+/** Calculates Iono and Tropo correction
+ * \param n_ready_tdcp The number of measurements written to `m_tdcp`
+ * \param nav_meas_tdcp Pointer to measurements array
+ * \param pos_ecef position in WGS84 Earth Centered, Earth Fixed (ECEF) Cartesian
+ * \param pos_llh position in WGS84 geodetic coordinates*/
+void calc_iono_tropo(u8 n_ready_tdcp,
```
ljbade: Im not sure if this is the best location for this function?

I think it might be more useful in libswiftnav

ljbade: Oh, but I see it needs the iono params struct... hmm I guess that would need to be made a parameter :-/

@benjamin0 Do you think it would be useful to have the correction function in libswiftnav?

src/solution.c
```diff
@@ -429,6 +434,56 @@ static void sol_thd_sleep(systime_t *deadline, systime_t interval)
...
+                                        pos_llh[0],
+                                        pos_llh[1],
+                                        az,el,
+                                        &iono_params);
+      /* correct pseudorange */
+      nav_meas_tdcp[i].pseudorange -= iono_correction;
+      /* correct carrier phase based on iono correction */
+      cp_correction = iono_correction * (GPS_L1_HZ + nav_meas_tdcp[i].doppler)
+                      / GPS_C;
+      nav_meas_tdcp[i].carrier_phase -= cp_correction;
```
ljbade: Can you also add the comment "sign here is opposite to normal due to
 +     * Piksi's unusual sign convention on carrier phase"

src/solution.c
```diff
@@ -429,6 +434,56 @@ static void sol_thd_sleep(systime_t *deadline, systime_t interval)
...
+    }
+    /* calculate tropo correction.
```
ljbade: Add a newline between comment and the bracket

src/solution.c
```diff
@@ -429,6 +434,56 @@ static void sol_thd_sleep(systime_t *deadline, systime_t interval)
...
+    wgsecef2azel(nav_meas_tdcp[i].sat_pos, pos_ecef, &az, &el);
+    /* read iono parameters for the tracking channel if available */
+    if (gps_iono_params_read(&iono_params)) {
+      /* calculate iono correction */
+      iono_correction = calc_ionosphere(&nav_meas_tdcp[i].tot,
+                                        pos_llh[0],
+                                        pos_llh[1],
+                                        az,el,
+                                        &iono_params);
+      /* correct pseudorange */
```
nsirola: Correct for signal frequency here (if not L1) with Leith's formula, something like:
iono_correction *= (this_frequency/L1_Hz)**2

src/solution.c
```diff
@@ -429,6 +434,56 @@ static void sol_thd_sleep(systime_t *deadline, systime_t interval)
...
+      nav_meas_tdcp[i].pseudorange -= iono_correction;
+      /* correct carrier phase based on iono correction */
+      cp_correction = iono_correction * (GPS_L1_HZ + nav_meas_tdcp[i].doppler)
```
nsirola: Use actual L1/L2 frequency here

src/solution.c
dt-exafore: Yes, we need to move this function to libswiftnav, further more we need to move the function to libswiftnav-private and update it for L2C using formula to convert from L1C Klobushar correction.
So, I'm going to create 2 new PRs for that.

src/solution.c
dt-exafore: Please, note:
This change depends on PR https://github.com/swift-nav/libswiftnav-private/pull/6 where following constants and functions defined: GPS_L2_HZ, code_to_carr_freq()

src/solution.c
dt-exafore: same like above

src/solution.c
dt-exafore: Please, find PR for public libswiftnav here https://github.com/swift-nav/libswiftnav/pull/362

src/solution.c
dt-exafore: Updated here https://github.com/swift-nav/libswiftnav-private/pull/32

rge-exafore:
moved to private (https://github.com/swift-nav/piksi_firmware_private/pull/16) and closed

PR #692, opened 2016-05-20
Seed RNG with sample data on Piksi v3.

gsmcmullin:
/cc @swift-nav/firmware

### DIFF
--- a/src/board/v3/init.c
+++ b/src/board/v3/init.c
@@ -23,8 +23,10 @@
 #include "peripherals/leds.h"
 #include "board/nap/nap_common.h"
 #include "nap/nap_conf.h"
+#include "nap/fft.h"
 #include "sbp.h"
 #include "error.h"
+#include "frontend.h"
 
 #define REQUIRED_NAP_VERSION_MASK (0xFFFF0000U)
 #define REQUIRED_NAP_VERSION_VAL  (0x03000000U)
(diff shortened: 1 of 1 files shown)


src/board/v3/init.c
```diff
@@ -78,7 +91,7 @@ void init(void)
   nap_auth_check();
   nap_callbacks_setup();
 
-  srand(0);
+  random_init();
```
jacobmcnamee: Need to initialize frontend before trying to read out samples.

src/board/v3/init.c
```diff
@@ -68,6 +69,18 @@ void pre_init(void)
   led_setup();
 }
 
+static void random_init(void)
+{
+  u64 sample_data;
+  u32 seed = 0;
+
+  raw_samples_get(&sample_data, sizeof(sample_data));
```
jacobmcnamee: Log error if `raw_samples_get()` returns false.

jacobmcnamee:
@wltr What bits of the samples are actually "random"? There is now a counter in the upper nibble right?

jacobmcnamee:
See #689

src/board/v3/init.c
gsmcmullin: This happens in `frontend_configure` called from `nap_setup` above.

wltr:
@jacobmcnamee The 4 LSBs are random. The counter was definitely needed for the bare-metal sample grabber firmware, but otherwise I don't think so. However, I suppose it can't hurt to be able to make sure that you get a continuous stream.

jacobmcnamee:
Can you delete https://github.com/swift-nav/piksi_firmware/blob/master/src/board/v3/frontend.c#L155 since `frontend_configure()` is now called in v3 `init()` ?

PR #695, opened 2016-05-20
Fix calc_sat_state() and calc_nav_meas() error checking and locking

jacobmcnamee:


ljbade:
@jacobmcnamee Can you rebase this?

@rge-exafore spotted the same problem

src/solution.c
```diff
@@ -490,23 +492,23 @@ static void solution_thread(void *arg)
...
+    s8 pvt_ret = calc_PVT(n_ready_tdcp, nav_meas_tdcp, disable_raim,
+                          &position_solution, &dops);
+    if (pvt_ret < 0) {
       /* An error occurred with calc_PVT! */
       /* TODO: Make this based on time since last error instead of a simple
        * count. */
       /* pvt_err_msg defined in libswiftnav/pvt.c */
       DO_EVERY((u32)soln_freq,
-        log_warn("PVT solver: %s (code %d)", pvt_err_msg[-ret-1], ret);
+        log_warn("PVT solver: %s (code %d)", pvt_err_msg[-pvt_ret-1], pvt_ret);
```
ljbade: At some stage it would be nice to have a similar error message lookup for the dgnss_* functions instead of returning numbers.

ljbade:
Looks good, but make sure to do a bunch of HITL runs

PR #701, opened 2016-05-26
New obs message and lock time (try 2)

denniszollo:


src/base_obs.c
```diff
@@ -417,6 +418,13 @@ void base_obs_setup()
     &deprecated_callback,
     &deprecated_node
   );
+  
+  static sbp_msg_callbacks_node_t deprecated_node_2;
+  sbp_register_cbk(
+    SBP_MSG_OBS_DEP_B,
```
mfine: Maybe drop `SBP_MSG_OBS_DEP_A` - and just put out a deprecated warning for `SBP_MSG_OBS_DEP_B`, not both.

src/track/track_gps_l1ca.c
```diff
@@ -140,7 +140,6 @@ static void tracker_gps_l1ca_init(const tracker_channel_info_t *channel_info,
 
   memset(data, 0, sizeof(gps_l1ca_tracker_data_t));
   tracker_ambiguity_unknown(channel_info->context);
-
```
jacobmcnamee: remove

src/solution.c
```diff
@@ -762,7 +762,7 @@ void process_matched_obs(u8 n_sds, gps_time_t *t, sdiff_t *sds, u16 base_id)
   }
 }
 
-static THD_WORKING_AREA(wa_time_matched_obs_thread, 20000);
+static THD_WORKING_AREA(wa_time_matched_obs_thread, 18000);
```
jacobmcnamee: Remove this change. The heap issue is likely not related to this PR, but we need to fix it separately and in a less hacky manner.

src/base_obs.c
denniszollo: disagree - we should send the warning no matter what.

mfine: After a deprecation period, we should be comfortable removing warnings. We did for the old unpacked obs messages. If it's not clear, I'm saying warn on `DEP_B`, not also `DEP_A` (note this is also generally a poor way to do "version" checking between devices).

src/sbp_utils.c
```diff
@@ -188,6 +188,30 @@ void sbp_make_heading(msg_baseline_heading_t *baseline_heading, const gps_time_t
...
+       for(u8 i = 0; i < 16; i++) {
+         if ((int)lock_time > (1<<(i+5))) {
+           continue;
+         }
+         else
+           return i;
+        }
+     }
+     return 15;
+}
```
mfine: Style is off on these two (indentation, curly braces). Logic could also be simplified a little bit:

```c
u32 min_lock_time_decode(u8 lock_time_packed) {
  if (lock_time_packed == 0)
    return 0;

  return 1 << (lock_time_packed + 4);
}

u8 lock_time_encode(u32 lock_time) {
  for(u8 i = 0; i < 16; i++)
    if (lock_time <= (u32)(1 << (i+5)))
      return i;

  return 15;
}
```

Would be good to add some unit tests for these as well.

denniszollo:
deprecated in favor of a soon to exist PR.

denniszollo:
superceded by #706

PR #704, opened 2016-05-26
Fix Travis Ubuntu ARM toolchain installation

pmiettinen:


jacobmcnamee:
FYI:

"For our 2015Q4 GCC 5 release we decided to move from the old PPA maintained by Terry Guo [1] to a team maintained one [2]. We also took advantage of that move to rename the package from gcc-arm-none-eabi to gcc-arm-embedded."

https://launchpad.net/gcc-arm-embedded/+announcement/13824

jacobmcnamee:
/cc @swift-nav/firmware @mookerji 

Any objections?

mookerji:
thanks! @pmiettinen !

PR #708, opened 2016-05-27
Change version to include build time and commit id of libswiftnav

rge-exafore:


src/Makefile
```diff
@@ -190,7 +190,19 @@ CPPWARN = -Wall -Wextra -Werror
...
+# Example: v0.21-05271044-4293c57-f82b1f4
+GIT_VERSION := $(shell git describe --abbrev=0)
+GIT_VERSION += $(shell date +%m%d%H%M)
+GIT_VERSION += $(shell git rev-parse --short=7 HEAD)
+GIT_VERSION += $(shell git -C ../libswiftnav rev-parse --short=7 HEAD)
+dash := -
+empty :=
+space := $(empty) $(empty)
+GIT_VERSION := $(subst $(space),$(dash),$(GIT_VERSION))
+$(info $(GIT_VERSION))
```
valeri-atamaniouk: Is it a debug statement? Some more information may be?

valeri-atamaniouk:
looks good

ljbade:
Looks OK to me, what do the new version string look like?

jacobmcnamee:
The submodule commit IDs are implicitly captured by the piksi_firmware commit ID. What is the motivation for including the libswiftnav commit ID explicitly?

rge-exafore:
libswiftnav changes more frequently than other submodules that's why including it explicitly. (But including all submodules' commit ids would be even better.) This ensures that the build can be restored even if submodule pointer wasn't moved or public version of libswiftnav was replaced with private or vice versa.

PR #710, opened 2016-05-31
Consolidate nap constants

axlan:
@jacobmcnamee @wltr

### DIFF
--- a/src/board/v3/nap/nap_constants.h
+++ b/src/board/v3/nap/nap_constants.h
@@ -13,11 +13,39 @@
 #ifndef SWIFTNAV_NAP_CONSTANTS_H
 #define SWIFTNAV_NAP_CONSTANTS_H
 
+/**
+ * @brief   The sampling rate of the samples coming off the frontend
+ */
 #define NAP_FRONTEND_SAMPLE_RATE_Hz                                (99.375e6)
 
+/**
+ * @brief   The sample rate decimation used by acquisition
+ */
 #define NAP_ACQ_DECIMATION_RATE                                          (12)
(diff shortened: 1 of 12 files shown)


src/board/v3/nap/nap_constants.h
```diff
@@ -13,11 +13,39 @@
...
 
+/**
+ * @brief   The sampling rate of the samples coming off the frontend
+ */
 #define NAP_FRONTEND_SAMPLE_RATE_Hz                                (99.375e6)
 
+/**
+ * @brief   The sample rate decimation used by acquisition
+ */
 #define NAP_ACQ_DECIMATION_RATE                                          (12)
```
wltr: The NAP has a register to tell you the decimation factor.

src/board/v3/nap/nap_constants.h
```diff
@@ -13,11 +13,39 @@
...
+ * @brief   The acquisition sample rate after decimation
+ * @note    This is the effective sampling rate of the acquisition results
+ */
+#define NAP_ACQ_SAMPLE_RATE_Hz             (NAP_FRONTEND_SAMPLE_RATE_Hz  \
+		                                            / NAP_ACQ_DECIMATION_RATE)
+
+/**
+ * @brief   The sample rate decimation used by tracking channels
+ */
+#define NAP_TRACK_DECIMATION_RATE                                         (4)
```
wltr: The NAP has a register to tell you the decimation factor.

axlan: I was preserving the original behavior. There may be some compile time optimization possible in having these values hard coded, but I agree it's probably better to check them at run time so they don't need to be kept in sync with any changes in the hardware. This is a bigger change though, should it be a separate pull request or does it make more sense to keep it in this one?

PR #711, opened 2016-05-31
Define GIT_VERSION in version.h

jacobmcnamee:
This allows source files dependent on `GIT_VERSION` to be rebuilt when `version.h` changes using standard file dependencies.


rge-exafore:
@jacobmcnamee could you please also review this change? https://github.com/swift-nav/piksi_firmware/pull/708 It adds time stamp and libswiftnav commit id to the version string.

PR #712, opened 2016-06-01
Use chprintf() for all printf() calls

jacobmcnamee:
Standard library `printf()` and `scanf()` are not thread-safe and use dynamic memory.

Requires https://github.com/swift-nav/ChibiOS/pull/7

TODO: Deal with `scanf()` calls.


jacobmcnamee:
Replaced by #714

PR #724, opened 2016-06-10
lock_detector: updated lock detector coefficients

valeri-atamaniouk:
Corrected coefficients for optimistic and pessimistic counters in lock detector.


ljbade:
Do you have HITL results?

@jacobmcnamee You understand the lock detector code?

